PR #1, opened 2022-05-13
Add x509 parsing utilities

rgnote:
* Moving x509 cert, key parsing utilities from notation-go to notation-core-go
* Will raise a separate PR for refactoring notation-go and notation repos

Signed-off-by: rgnote <5878554+rgnote@users.noreply.github.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module github.com/notaryproject/notation-core-go
+
+go 1.18
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,3 @@
 module github.com/notaryproject/notation-core-go
 
-go 1.18
+go 1.17
--- a/x509/cert.go
+++ b/x509/cert.go
@@ -1,0 +1,31 @@
+package cryptoutil
+
+import (
+	"crypto/x509"
+	"encoding/pem"
+	"os"
+)
+
+// ReadCertificateFile reads a certificate PEM file.
+func ReadCertificateFile(path string) ([]*x509.Certificate, error) {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	return ParseCertificatePEM(data)
+}
+
+// ParseCertificatePEM parses a certificate PEM.
+func ParseCertificatePEM(data []byte) ([]*x509.Certificate, error) {
+	var certs []*x509.Certificate
+	block, rest := pem.Decode(data)
+	for block != nil {
+		cert, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			return nil, err
+		}
+		certs = append(certs, cert)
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/notaryproject/notation-core-go
+
+go 1.18
```
rgnote, line 3: I want someone with more familiarity with golang chime in here. Currently, notation and notation-go are on 1.17. Reading 1.18 [release notes](https://go.dev/doc/go1.18), it looks like 1.18 is backward compatible and it is also fixing some bugs present in earlier Go versions. Let me know if there a reason we shouldn't upgrade to 1.18.

gokarnm: @shizhMSFT , @qmuntal, any feedback here?

gokarnm: @shizhMSFT , @qmuntal, any feedback here?

qmuntal: The `go` directive inside `go.mod` does not refer to the go version that will be used to compile the library, but to the maximum set of language features that this library supports. For example, generics are only available if we set it to `1.18` and `embed` if we set it to `1.16` or greater.

A customer using a lower Go toolchain version can still use a module whose `go` directive if set to a higher version unless the module really uses a newer language feature.

The best practice is to set the `go` directive to the lowest supported Go version. As we are not using generics here, I would set it to `1.17` to be in-line with `notation-go`.

rgnote: Good to know. Updated the PR.

@gokarnm Can you merge if this looks good?

x509/cert.go
```diff
@@ -0,0 +1,31 @@
...
+func ReadCertificateFile(path string) ([]*x509.Certificate, error) {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	return ParseCertificatePEM(data)
+}
+
+// ParseCertificatePEM parses a certificate PEM.
+func ParseCertificatePEM(data []byte) ([]*x509.Certificate, error) {
```
gokarnm, line 19:
```suggestion
// ParseCertificatePEM parses a certificate PEM,
// Does not perform any additional validations if the certs are a valid cert chain.
func ParseCertificatePEM(data []byte) ([]*x509.Certificate, error) {
```

gokarnm (approved):

dtzar:
LGTM

jondonas (approved):

qmuntal (approved):

dtzar (approved):

NiazFK (approved):
LGTM

NiazFK:
Add x509 parsing utilities

Signed-off-by: rgnote <5878554+rgnote@users.noreply.github.com>
Co-authored-by: Milind Gokarn <milind81@gmail.com>

PR #2, opened 2022-05-19
Add method to verify certificate chain

jondonas:
Signed-off-by: Jonathan Donas <jddonas@amazon.com>


x509/cert.go
```diff
@@ -30,3 +31,30 @@ func ParseCertificatePEM(data []byte) ([]*x509.Certificate, error) {
...
+	}
+
+	for i, cert := range certChain {
+		if i == len(certChain)-1 {
+			if !isRootCertificate(cert) {
+				return errors.New("certificate chain does not end in a root")
+			}
+		} else {
+			if err := cert.CheckSignatureFrom(certChain[i+1]); err != nil || isRootCertificate(cert) {
+				return errors.New("certificate chain does not validate sequentially from signing cert to root")
```
rgnote, line 49: This error message can be bit more descriptive about what failed in the validation. 

```
signature on certificate "<subject>" is not issued by "<subject>"
```

x509/cert.go
```diff
@@ -30,3 +31,30 @@ func ParseCertificatePEM(data []byte) ([]*x509.Certificate, error) {
 	}
 	return certs, nil
 }
+
+// Validates a certificate chain from leaf to root. Expects certChain input to be ordered
+// Returns the full ordered chain: [leaf, intermediates..., root]
+func buildAndValidateCertChain(certChain []*x509.Certificate) error {
+	if len(certChain) < 2 {
+		return errors.New("certificate chain must contain at least two certificates")
```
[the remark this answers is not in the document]
jondonas: One of the constraints for a valid chain, specified [here](https://hackmd.io/1ScKx83xTfuabvYaaQhPhw) under "Other Constraints"

A valid certificate chain MUST contain a minimum of two certificates - a leaf and a root

x509/cert.go
```diff
@@ -30,3 +31,35 @@ func ParseCertificatePEM(data []byte) ([]*x509.Certificate, error) {
...
+	return nil
+}
+
+func isSelfSigned(cert *x509.Certificate) bool {
+	return isIssuedBy(cert, cert)
+}
+
+func isIssuedBy(subject *x509.Certificate, issuer *x509.Certificate) bool {
+	err := subject.CheckSignatureFrom(issuer)
+	return err == nil && issuer.Subject.String() == subject.Issuer.String()
```
rgnote, line 64: DNs are structures and shouldn't be treated as strings. See https://frasertweedale.github.io/blog-redhat/posts/2019-05-28-a-dn-is-not-a-string.html

Can we use Golang's in-built [chain validation](https://github.com/golang/go/blob/master/src/crypto/x509/verify.go#L749)?

x509/cert.go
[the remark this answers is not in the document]
rgnote: @gokarnm @priteshbandi
It is not clear why we are not supporting self-signed certificates, which may be easier for testing purposes.

x509/cert.go
jondonas: I actually originally wrote chain validation using Go's Verify method, but it's not ideal because it's intended for TLS certs. The code ends up being [2-3 times longer](https://paste.amazon.com/show/jddonas/1653497401) than what I have here, you have to specify a "current time" for the certs even if you want expired certs to still validate, and you have to know ahead of time the exact roots you want to verify against (you can't just validate up to "any" root)

In terms of the DN, I'll look into a better way to check equality. My original reasoning is that checking string equality should at least be consistent because I'm not sure how consistent struct equality is in Go. Will get back about this

x509/cert.go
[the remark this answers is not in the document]
iamsamirzon: @rgnote - I too think the implementation should support self signed certificates. The most common use case will be testing. We should also document ( or link to existing document) how someone can generate a self signed certificate.

x509/cert.go
jondonas: Fixed this by copying how the Verify method checks this - comparing raw Subject bytes to Issues bytes: [source](https://github.com/golang/go/blob/master/src/crypto/x509/verify.go#L565)

x509/cert.go
```diff
@@ -30,3 +32,35 @@ func ParseCertificatePEM(data []byte) ([]*x509.Certificate, error) {
...
+	}
+
+	for i, cert := range certChain {
+		if i == len(certChain)-1 {
+			if !isSelfSigned(cert) {
+				return errors.New("certificate chain must end with a root certificate (root certificates are self-signed)")
+			}
+		} else {
+			if isSelfSigned(cert) {
+				return errors.New("certificate chain must not contain self-signed intermediates")
```
rgnote, line 49:
```suggestion
				return errors.New("certificate chain must not contain self-signed intermediate certificates")
```

x509/cert.go
```diff
@@ -30,3 +32,35 @@ func ParseCertificatePEM(data []byte) ([]*x509.Certificate, error) {
...
+			} else if nextCert := certChain[i+1]; !isIssuedBy(cert, nextCert) {
+				return errors.New("signature on certificate '" + nextCert.Subject.String() + "' is not issued by '" + cert.Issuer.String() + "'")
```
rgnote, line 51:
```suggestion
				return fmt.Errorf("signature on certificate %q is not issued by %q", cert.Subject.String(), nextCert.Subject.String())
```

rgnote (changes requested):

rgnote (approved):

x509/cert.go
```diff
@@ -30,3 +33,35 @@ func ParseCertificatePEM(data []byte) ([]*x509.Certificate, error) {
 	}
 	return certs, nil
 }
+
+// ValidateCertChain takes an ordered certificate chain and validates issuance from leaf to root
+func ValidateCertChain(certChain []*x509.Certificate) error {
```
gokarnm, line 38: This method only check if the cert chain is correctly formed, not that it chains to an trusted root certificate in trust store correct? Where will we have that logic. Also is this the method where we will add additional cert chain constraint check (KU, EKU, path len)?

We should create an issue to track that work and add TODO with the issue link, if that logic will be included here.

gokarnm, line 38: I'd rename this to `ValidateSigningCertChain` if this is going to be distinct from TSA cert chain validation.

gokarnm (changes requested):

PR #5, opened 2022-06-02
moving TSA timestamping and internal crypto utils from notation-go

rgnote:
No change other than moving the timestamping code and it's dependencies from notation-go to notation-core-go

Signed-off-by: rgnote <5878554+rgnote@users.noreply.github.com>


shizhMSFT (approved):
LGTM

gokarnm (approved):
Approving the refactor which moved code as is from notation-go. We still need to make a pass over this code and see if there are alternatives to writing the CMS parsing and related ASN.1 code.

PR #7, opened 2022-06-14
Abstract out signature generation and verification to SignatureEnvelope

priteshbandi:
Change Log:
- Implements SignatureEnvelope struct that would be used by notation-go-lib to generate/verify the signature.
- SignatureEnvelope internally implements actual signature formats like jws(cose in future)
- Implemented JWS-JSON signature generation and verification code.

### DIFF
--- a/internal/testhelper/certificatetest.go
+++ b/internal/testhelper/certificatetest.go
@@ -1,0 +1,149 @@
+// Package testhelper implements utility routines required for writing unit tests.
+// The testhelper should only be used in unit tests.
+package testhelper
+
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"math/big"
+	"time"
+)
+
+var (
+	rsaRoot     CertTuple
+	rsaLeaf     CertTuple
+	ecdsaRoot   ECCertTuple
+	ecdsaLeaf   ECCertTuple
+	unsupported CertTuple
+)
+
+type CertTuple struct {
+	Cert       *x509.Certificate
+	PrivateKey *rsa.PrivateKey
+}
+
+type ECCertTuple struct {
+	Cert       *x509.Certificate
+	PrivateKey *ecdsa.PrivateKey
+}
+
+// init runs before any other part of this package.
+func init() {
+	setupCertificates()
+}
+
+// GetRSARootCertificate returns root certificate signed using RSA algorithm
+func GetRSARootCertificate() CertTuple {
+	return rsaRoot
+}
+
+// GetRSALeafCertificate returns leaf certificate signed using RSA algorithm
+func GetRSALeafCertificate() CertTuple {
+	return rsaLeaf
+}
+
+// GetECRootCertificate returns root certificate signed using EC algorithm
+func GetECRootCertificate() ECCertTuple {
+	return ecdsaRoot
+}
+
+// GetECLeafCertificate returns leaf certificate signed using EC algorithm
+func GetECLeafCertificate() ECCertTuple {
+	return ecdsaLeaf
+}
+
+// GetUnsupportedCertificate returns certificate signed using RSA algorithm with key size of 1024 bits
+// which is not supported by notary.
+func GetUnsupportedCertificate() CertTuple {
+	return unsupported
+}
+
+func setupCertificates() {
+	rsaRoot = getCertTuple("Notation Test Root", nil)
+	rsaLeaf = getCertTuple("Notation Test Leaf Cert", &rsaRoot)
+	ecdsaRoot = getECCertTuple("Notation Test Root2", nil)
+	ecdsaLeaf = getECCertTuple("Notation Test Leaf Cert", &ecdsaRoot)
+
+	// This will be flagged by the static code analyzer as 'Use of a weak cryptographic key' but its intentional
+	// and is used only for testing.
+	k, _ := rsa.GenerateKey(rand.Reader, 1024)
+	unsupported = getRSACertTupleWithPK(k, "Notation Unsupported Root", nil)
+}
+
+func getCertTuple(cn string, issuer *CertTuple) CertTuple {
+	pk, _ := rsa.GenerateKey(rand.Reader, 3072)
+	return getRSACertTupleWithPK(pk, cn, issuer)
+}
+
+func getECCertTuple(cn string, issuer *ECCertTuple) ECCertTuple {
+	k, _ := ecdsa.GenerateKey(elliptic.P384(), rand.Reader)
+	return getECDSACertTupleWithPK(k, cn, issuer)
+}
+
+func getRSACertTupleWithPK(privKey *rsa.PrivateKey, cn string, issuer *CertTuple) CertTuple {
+	template := getCertTemplate(issuer == nil, cn)
+
+	var certBytes []byte
+	if issuer != nil {
+		certBytes, _ = x509.CreateCertificate(rand.Reader, template, issuer.Cert, &privKey.PublicKey, issuer.PrivateKey)
+	} else {
+		certBytes, _ = x509.CreateCertificate(rand.Reader, template, template, &privKey.PublicKey, privKey)
+	}
+
+	cert, _ := x509.ParseCertificate(certBytes)
+	return CertTuple{
+		Cert:       cert,
+		PrivateKey: privKey,
+	}
+}
+
+func getECDSACertTupleWithPK(privKey *ecdsa.PrivateKey, cn string, issuer *ECCertTuple) ECCertTuple {
+	template := getCertTemplate(issuer == nil, cn)
+
+	var certBytes []byte
+	if issuer != nil {
+		certBytes, _ = x509.CreateCertificate(rand.Reader, template, issuer.Cert, &privKey.PublicKey, issuer.PrivateKey)
+	} else {
+		certBytes, _ = x509.CreateCertificate(rand.Reader, template, template, &privKey.PublicKey, privKey)
+	}
+
+	cert, _ := x509.ParseCertificate(certBytes)
+	return ECCertTuple{
+		Cert:       cert,
+		PrivateKey: privKey,
+	}
+}
+
+func getCertTemplate(isRoot bool, cn string) *x509.Certificate {
+	template := &x509.Certificate{
+		Subject: pkix.Name{
+			Organization: []string{"Notary"},
+			Country:      []string{"US"},
+			Province:     []string{"WA"},
+			Locality:     []string{"Seattle"},
+			CommonName:   cn,
+		},
+		NotBefore:   time.Now(),
+		KeyUsage:    x509.KeyUsageDigitalSignature,
+		ExtKeyUsage: []x509.ExtKeyUsage{x509.ExtKeyUsageCodeSigning},
+	}
+	if isRoot {
+		template.SerialNumber = big.NewInt(1)
+		template.NotAfter = time.Now().AddDate(0, 1, 0)
+		template.KeyUsage = x509.KeyUsageCertSign
+		template.BasicConstraintsValid = true
+		template.MaxPathLen = 1
+		template.IsCA = true
+	} else {
+		template.SerialNumber = big.NewInt(2)
+		template.NotAfter = time.Now().AddDate(0, 0, 1)
+		template.KeyUsage = x509.KeyUsageDigitalSignature
+		template.ExtKeyUsage = []x509.ExtKeyUsage{x509.ExtKeyUsageCodeSigning}
+	}
+
+	return template
+}
--- a/signer/errors.go
+++ b/signer/errors.go
@@ -1,0 +1,103 @@
+package signer
+
+import "fmt"
+
+// SignatureIntegrityError is used when the Signature associated is no longer valid.
+type SignatureIntegrityError struct {
+	err error
+}
+
+func (e SignatureIntegrityError) Error() string {
+	return fmt.Sprintf("signature is invalid. Error: %s", e.err.Error())
+}
+
+// MalformedSignatureError is used when Signature envelope is malformed.
+type MalformedSignatureError struct {
+	msg string
+}
+
+func (e MalformedSignatureError) Error() string {
+	if e.msg != "" {
+		return e.msg
+	}
+	return "signature envelope format is malformed"
+
+}
+
+// UnsupportedSignatureFormatError is used when Signature envelope is not supported.
+type UnsupportedSignatureFormatError struct {
+	mediaType string
+}
+
+func (e UnsupportedSignatureFormatError) Error() string {
+	return fmt.Sprintf("signature envelope format with media type %q is not supported", e.mediaType)
+}
+
+// SignatureNotFoundError is used when signature envelope is not present.
+type SignatureNotFoundError struct{}
+
+func (e SignatureNotFoundError) Error() string {
+	return "signature envelope is not present"
+}
+
+// SignatureAuthenticityError is used when signature is not generated using trusted certificates.
+type SignatureAuthenticityError struct{}
+
+func (e SignatureAuthenticityError) Error() string {
+	return "signature is not produced by a trusted signer"
+}
+
+// UnsupportedOperationError is used when an operation is not supported.
+type UnsupportedOperationError struct {
+	operation string
+}
+
+func (e UnsupportedOperationError) Error() string {
+	return fmt.Sprintf("%q operation is not supported", e.operation)
+}
+
+// UnsupportedSigningKeyError is used when a signing key is not supported
+type UnsupportedSigningKeyError struct {
+	keyType string
+}
+
+func (e UnsupportedSigningKeyError) Error() string {
+	if e.keyType != "" {
+		return fmt.Sprintf("%q signing key is not supported", e.keyType)
+	}
+	return "signing key is not supported"
+}
+
+// MalformedArgumentError is used when an argument to a function is malformed.
+type MalformedArgumentError struct {
+	param string
+	err   error
+}
+
+func (e MalformedArgumentError) Error() string {
+	if e.err != nil {
+		return fmt.Sprintf("%q param is malformed. Error: %s", e.param, e.err.Error())
+	}
+	return fmt.Sprintf("%q param is malformed", e.param)
+}
+
+// MalformedSignRequestError is used when SignRequest is malformed.
+type MalformedSignRequestError struct {
+	msg string
+}
+
+func (e MalformedSignRequestError) Error() string {
+	if e.msg != "" {
+		return e.msg
+	}
+	return "SignRequest is malformed"
+}
+
+// SignatureAlgoNotSupportedError is used when signing algo is not supported.
+type SignatureAlgoNotSupportedError struct {
+	alg string
+}
+
+func (e SignatureAlgoNotSupportedError) Error() string {
+	return fmt.Sprintf("signature algorithm %q is not supported", e.alg)
+}
--- a/signer/jws.go
+++ b/signer/jws.go
@@ -1,0 +1,392 @@
+package signer
+
+import (
+	"crypto/x509"
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/golang-jwt/jwt/v4"
+)
+
+const (
+	MediaTypeJWSJson SignatureMediaType = "application/jose+json"
+)
+
+const (
+	headerKeyExpiry      = "io.cncf.notary.expiry"
+	headerKeySigningTime = "io.cncf.notary.signingTime"
+	headerKeyCrit        = "crit"
+	headerKeyAlg         = "alg"
+	headerKeyCty         = "cty"
+)
+
+// jwsEnvelope represents implements internalSignatureEnvelope interface.
+type jwsEnvelope struct {
+	internalEnv *jwsInternalEnvelope
+}
+
+func newJWSEnvelopeFromBytes(envelopeBytes []byte) (*jwsEnvelope, error) {
+	jwsInternal, err := newJwsInternalEnvelopeFromBytes(envelopeBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	return &jwsEnvelope{internalEnv: jwsInternal}, nil
+}
+
+func (jws *jwsEnvelope) validateIntegrity() error {
+	if jws.internalEnv == nil {
+		return SignatureNotFoundError{}
+	}
+
+	if len(jws.internalEnv.Header.CertChain) == 0 {
+		return MalformedSignatureError{msg: "malformed leaf certificate"}
+	}
+
+	cert, err := x509.ParseCertificate(jws.internalEnv.Header.CertChain[0])
+	if err != nil {
+		return MalformedSignatureError{msg: "malformed leaf certificate"}
+	}
+
+	// verify JWT
+	compact := strings.Join([]string{jws.internalEnv.Protected, jws.internalEnv.Payload, jws.internalEnv.Signature}, ".")
+	return verifyJWT(compact, cert.PublicKey)
+}
+
+func (jws *jwsEnvelope) signPayload(req SignRequest) ([]byte, error) {
+	leafPublicKey := req.CertificateChain[0].PublicKey
+	var m map[string]interface{}
+	if err := json.Unmarshal(req.Payload, &m); err != nil {
+		return nil, err
+	}
+	signingMethod, err := getSigningMethod(leafPublicKey)
+	if err != nil {
+		return nil, err
+	}
+
+	signedAttrs, err := getSignedAttrs(req, signingMethod)
+	if err != nil {
+		return nil, err
+	}
+
+	compact, err := signJWT(m, signedAttrs, signingMethod, req.SignatureProvider)
+	if err != nil {
+		return nil, err
+	}
+
+	j, err := generateJws(compact, req)
+	if err != nil {
+		return nil, err
+	}
+
+	b, err := json.Marshal(j)
+	if err != nil {
+		return nil, err
+	}
+	jws.internalEnv = j
+	return b, nil
+}
+
+func (jws *jwsEnvelope) getSignerInfo() (*SignerInfo, error) {
+	signInfo := SignerInfo{}
+	if jws.internalEnv == nil {
+		return nil, SignatureNotFoundError{}
+	}
+
+	// parse payload
+	payload, err := base64.RawURLEncoding.DecodeString(jws.internalEnv.Payload)
+	if err != nil {
+		return nil, err
+	}
+	signInfo.Payload = payload
+
+	// parse protected headers
+	protected, err := parseProtectedHeaders(jws.internalEnv.Protected)
+	if err != nil {
+		return nil, err
+	}
+	if err := populateProtectedHeaders(protected, &signInfo); err != nil {
+		return nil, err
+	}
+
+	// parse signature
+	sig, err := base64.RawURLEncoding.DecodeString(jws.internalEnv.Signature)
+	if err != nil {
+		return nil, err
+	}
+	signInfo.Signature = sig
+
+	// parse headers
+	var certs []*x509.Certificate
+	for _, certBytes := range jws.internalEnv.Header.CertChain {
+		cert, err := x509.ParseCertificate(certBytes)
+		if err != nil {
+			return nil, err
+		}
+		certs = append(certs, cert)
+	}
+	signInfo.CertificateChain = certs
+	signInfo.UnsignedAttributes.SigningAgent = jws.internalEnv.Header.SigningAgent
+	signInfo.TimestampSignature = jws.internalEnv.Header.TimestampSignature
+
+	return &signInfo, nil
+}
+
+func parseProtectedHeaders(encoded string) (*jwsProtectedHeader, error) {
+	rawProtected, err := base64.RawURLEncoding.DecodeString(encoded)
+	if err != nil {
+		return nil, MalformedSignatureError{msg: fmt.Sprintf("jws envelope protected header can't be decoded: %s", err.Error())}
+	}
+
+	var protected jwsProtectedHeader
+	if err = json.Unmarshal(rawProtected, &protected); err != nil {
+		return nil, MalformedSignatureError{msg: fmt.Sprintf("jws envelope protected header can't be decoded: %s", err.Error())}
+	}
+	if err = json.Unmarshal(rawProtected, &protected.ExtendedAttributes); err != nil {
+		return nil, MalformedSignatureError{msg: fmt.Sprintf("jws envelope protected header can't be decoded: %s", err.Error())}
+	}
+
+	// delete attributes that are already defined in jwsProtectedHeader.
+	delete(protected.ExtendedAttributes, headerKeyAlg)
+	delete(protected.ExtendedAttributes, headerKeyCty)
+	delete(protected.ExtendedAttributes, headerKeyCrit)
+	delete(protected.ExtendedAttributes, headerKeySigningTime)
+	delete(protected.ExtendedAttributes, headerKeyExpiry)
+
+	return &protected, nil
+}
+
+func populateProtectedHeaders(pHeader *jwsProtectedHeader, signInfo *SignerInfo) error {
+	err := validateCriticalHeaders(pHeader)
+	if err != nil {
+		return err
+	}
+
+	if signInfo.SignatureAlgorithm, err = getAlgo(pHeader.Algorithm); err != nil {
+		return err
+	}
+
+	signInfo.PayloadContentType = pHeader.ContentType
+	signInfo.SignedAttributes.SigningTime = pHeader.SigningTime.Truncate(time.Second)
+	signInfo.SignedAttributes.Expiry = pHeader.Expiry.Truncate(time.Second)
+	signInfo.SignedAttributes.ExtendedAttributes = getExtendedAttributes(pHeader.ExtendedAttributes, pHeader.Critical)
+	return nil
+}
+
+func getExtendedAttributes(attrs map[string]interface{}, critical []string) []Attribute {
+	extendedAttr := make([]Attribute, 0, len(attrs))
+	for key, value := range attrs {
+		extendedAttr = append(extendedAttr, Attribute{
+			Key:      key,
+			Critical: contains(critical, key),
+			Value:    value,
+		})
+	}
+	return extendedAttr
+}
+
+func validateCriticalHeaders(pheader *jwsProtectedHeader) error {
+	if len(pheader.Critical) == 0 {
+		return MalformedSignatureError{"missing `crit` header"}
+	}
+
+	mustMarkedCrit := map[string]bool{}
+	if !pheader.Expiry.IsZero() {
+		mustMarkedCrit[headerKeyExpiry] = true
+	}
+
+	for _, val := range pheader.Critical {
+		if _, ok := mustMarkedCrit[val]; ok {
+			delete(mustMarkedCrit, val)
+		} else {
+			if _, ok := pheader.ExtendedAttributes[val]; !ok {
+				return MalformedSignatureError{msg: fmt.Sprintf("%q header is marked critical but not present", val)}
+			}
+		}
+	}
+
+	// validate all required critical headers are present.
+	if len(mustMarkedCrit) != 0 {
+		// This is not taken care by VerifySignerInfo method
+		return MalformedSignatureError{"required headers not marked critical"}
+	}
+
+	return nil
+}
+
+func getSignedAttrs(req SignRequest, method jwt.SigningMethod) (map[string]interface{}, error) {
+	extAttrs := make(map[string]interface{})
+	var crit []string
+	if !req.Expiry.IsZero() {
+		crit = append(crit, headerKeyExpiry)
+	}
+
+	for _, elm := range req.ExtendedSignedAttrs {
+		extAttrs[elm.Key] = elm.Value
+		if elm.Critical {
+			crit = append(crit, elm.Key)
+		}
+	}
+
+	pHeader := jwsProtectedHeader{
+		Algorithm:   method.Alg(),
+		ContentType: req.PayloadContentType,
+		Critical:    crit,
+		SigningTime: req.SigningTime.Truncate(time.Second),
+	}
+	if !req.Expiry.IsZero() {
+		pHeader.Expiry = req.Expiry.Truncate(time.Second)
+	}
+
+	m, err := convertToMap(pHeader)
+	if err != nil {
+		return nil, MalformedSignRequestError{msg: fmt.Sprintf("unexpected error occured while creating protected headers, Error: %s", err.Error())}
+	}
+
+	return mergeMaps(m, extAttrs), nil
+}
+
+// ***********************************************************************
+// jwsEnvelope-JSON specific code
+// ***********************************************************************
+const (
+	// PayloadContentTypeJWSV1 describes the media type of the jwsEnvelope envelope.
+	PayloadContentTypeJWSV1 = "application/vnd.cncf.notary.v2.jws.v1"
+)
+
+// jwsInternalEnvelope is the final Signature envelope.
+type jwsInternalEnvelope struct {
+	// JWSPayload Base64URL-encoded.
+	Payload string `json:"payload"`
+
+	// jwsProtectedHeader Base64URL-encoded.
+
+	Protected string `json:"protected"`
+
+	// Signature metadata that is not integrity Protected
+	Header jwsUnprotectedHeader `json:"header"`
+
+	// Base64URL-encoded Signature.
+	Signature string `json:"signature"`
+}
+
+// jwsProtectedHeader contains the set of protected headers.
+type jwsProtectedHeader struct {
+	// Defines which algorithm was used to generate the signature.
+	Algorithm string `json:"alg"`
+
+	// Media type of the secured content (the payload).
+	ContentType string `json:"cty"`
+
+	// Lists the headers that implementation MUST understand and process.
+	Critical []string `json:"crit"`
+
+	// The time at which the signature was generated.
+	SigningTime time.Time `json:"io.cncf.notary.signingTime"`
+
+	// The "best by use" time for the artifact, as defined by the signer.
+	Expiry time.Time `json:"io.cncf.notary.expiry,omitempty"`
+
+	// The user defined attributes.
+	ExtendedAttributes map[string]interface{} `json:"-"`
+}
+
+// jwsUnprotectedHeader contains the set of unprotected headers.
+type jwsUnprotectedHeader struct {
+	// RFC3161 time stamp token Base64-encoded.
+	TimestampSignature []byte `json:"io.cncf.notary.TimestampSignature,omitempty"`
+
+	// List of X.509 Base64-DER-encoded certificates
+	// as defined at https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.6.
+	CertChain [][]byte `json:"x5c"`
+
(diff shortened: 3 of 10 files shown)


signer/jws.go
```diff
@@ -0,0 +1,475 @@
...
+}
+
+func populateString(data map[string]interface{}, s string, holder *string) error {
+	if val, ok := data[s]; ok {
+		valString, ok := val.(string)
+		if !ok {
+			return MalformedSignatureError{msg: fmt.Sprintf("failed to parse %q in protected headers", s)}
+		}
+		*holder = valString
+		delete(data, s)
```
rgnote, line 177: Can you educate me why we need to delete this?

signer/jws.go
```diff
@@ -0,0 +1,475 @@
...
+	}
+	return nil
+}
+
+func populateAlg(data map[string]interface{}, holder *signatureAlgorithm) error {
+	var alg string
+	if err := populateString(data, "alg", &alg); err != nil {
+		return err
```
rgnote, line 185: Do we need to wrap these errors into `MalformedSignatureError`?

rgnote (changes requested):

signer/signer.go
```diff
@@ -0,0 +1,243 @@
...
+	// validateIntegrity validates the integrity of given Signature envelope.
+	validateIntegrity() error
+	// getSignerInfo returns the information stored in the Signature envelope and doesn't perform integrity verification.
+	getSignerInfo() (SignerInfo, error)
+	// signPayload created Signature envelope.
+	signPayload(SignRequest) ([]byte, error)
+}
+
+// Verify performs integrity validation and validates that cert-chain stored in Signature leads to given set of trusted root.
+func (s *SignatureEnvelope) Verify(trustedCerts []x509.Certificate) (x509.Certificate, error) {
```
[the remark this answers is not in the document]
priteshbandi: Its returning the certificate that was used for establishing trust i.e. authenticity of the signature.

signer/signer.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+// Sign generates Signature using given SignRequest.
+func (s *SignatureEnvelope) Sign(req SignRequest) ([]byte, error) {
+	if err := validateSignRequest(req); err != nil {
+		return []byte{}, err
+	}
+
+	sig, err := s.internalEnvelope.signPayload(req)
+	if sig != nil {
```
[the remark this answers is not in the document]
priteshbandi: Yep thats correct, updated the code.

signer/jws.go
```diff
@@ -0,0 +1,475 @@
...
+	// parse headers
+	certs := make([]x509.Certificate, 0)
+	for _, certBytes := range jws.internalEnv.Header.CertChain {
+		if cert, err := x509.ParseCertificate(certBytes); err != nil {
+			return signInfo, err
+		} else {
+			certs = append(certs, *cert)
+		}
+	}
+	signInfo.CertificateChain = certs
```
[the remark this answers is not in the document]
priteshbandi: This should be implemented as part of signer#verify method, replaced todo there with call to ValidateCertChain method.

signer/jws.go
```diff
@@ -0,0 +1,475 @@
...
+	if err := json.Unmarshal(req.Payload, &m); err != nil {
+		return []byte{}, err
+	}
+	signingMethod, err := getSigningMethod(leafPublicKey)
+	if err != nil {
+		return nil, err
+	}
+
+	signedAttrs := getSignedAttrs(req, signingMethod)
+	compact, _ := signJWT(m, signedAttrs, signingMethod, req.SignatureProvider)
```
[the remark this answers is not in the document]
priteshbandi: Yep, will add

signer/jws.go
```diff
@@ -0,0 +1,475 @@
...
+	return compact, nil
+}
+
+// verifyJWT verifies the JWT token against the specified verification key
+func verifyJWT(tokenString string, key crypto.PublicKey) error {
+	signingMethod, _ := getSigningMethod(key)
+
+	// parse and verify token
+	parser := &jwt.Parser{
+		ValidMethods:         []string{"PS256", "PS384", "PS512", "ES256", "ES384", "ES512"},
```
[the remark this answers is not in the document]
priteshbandi: In go we dont have array constants. I can create a variable but since its only used at one place will keep it here.

signer/signer.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	if req.SignatureProvider == nil {
+		return MalformedSignRequestError{msg: "SignatureProvider is nil"}
+	}
+	return nil
+}
+
+// NewSignatureEnvelopeFromBytes is used for signature verification workflow
+func NewSignatureEnvelopeFromBytes(envelopeBytes []byte, envelopeMediaType signatureMediaType) (SignatureEnvelope, error) {
```
shizhMSFT, line 215: Since `NewSignatureEnvelopeFromBytes` is an exported function, I expect `signatureMediaType` to be exported too.

shizhMSFT, line 215: According to the context, `SignatureEnvelope` should be an interface.

signer/jws.go
```diff
@@ -0,0 +1,478 @@
...
+	"github.com/golang-jwt/jwt/v4"
+)
+
+const (
+	JWS_JSON_MEDIA_TYPE signatureMediaType = "application/jose+json"
+)
+
+const (
+	expiryHeaderKey      = "io.cncf.notary.expiry"
+	signingTimeHeaderKey = "io.cncf.notary.signingTime"
```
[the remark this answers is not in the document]
priteshbandi: Its `signingTime` 
Ref: https://github.com/notaryproject/notaryproject/blob/main/signature-envelope-jws.md#protected-headers

signer/signer.go
priteshbandi: Done

signer/errors.go
```diff
@@ -0,0 +1,106 @@
+package signer
+
+import "fmt"
+
+// InvalidSignatureError is used when the Signature associated is no longer valid.
+type InvalidSignatureError struct {
```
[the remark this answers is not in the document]
priteshbandi: Looks like in go both formats `ErrInvalidSignature`  and `ErrorInvalidSignature` are equally valid.
https://cs.opensource.google/search?q=Err&ss=go%2Fgo&start=31

signer/jws.go
priteshbandi: So how we are determining`ExtendedAttributes` is by first fetching and then removing the known attributes. Once all known attributes are removed then whatever is reaming is considered as ExtendedAttributes

signer/signer.go
```diff
@@ -0,0 +1,243 @@
...
+
+	sig, err := s.internalEnvelope.signPayload(req)
+	if sig != nil {
+		s.rawSignatureEnvelope = sig
+	}
+	return sig, err
+}
+
+// GetSignerInfo returns information about the Signature envelope
+func (s SignatureEnvelope) GetSignerInfo() (SignerInfo, error) {
```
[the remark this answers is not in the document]
priteshbandi: Done

signer/jws.go
```diff
@@ -0,0 +1,475 @@
...
+// signJWT signs the given payload and headers using the given signing method and signature provider
+func signJWT(payload map[string]interface{}, headers map[string]interface{}, method jwt.SigningMethod, sigPro SignatureProvider) (string, error) {
+	var claims jwt.MapClaims = payload
+	token := &jwt.Token{
+		Header: headers,
+		Claims: claims,
+	}
+
+	token.Method = SigningMethodForSign{algo: method.Alg(), sigProvider: sigPro}
+	compact, err := token.SignedString("DummyNotUsed")
```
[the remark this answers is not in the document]
priteshbandi: Added explanation.

signer/signer.go
```diff
@@ -0,0 +1,243 @@
...
+	Sign([]byte) ([]byte, error)
+}
+
+type SignatureEnvelope struct {
+	rawSignatureEnvelope []byte
+	internalEnvelope     internalSignatureEnvelope
+}
+
+// Contains a set of common methods that every Signature envelope format must implement.
+type internalSignatureEnvelope interface {
```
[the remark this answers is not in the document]
priteshbandi: I think given small codebase it might not be worth moving it to internal pkg but if in future we add more sig format we can do this refactoring..

signer/signer.go
priteshbandi: It cannot be an interface as we are storing state(the signature).

signer/jws.go
priteshbandi: Not required as its taken care by https://github.com/priteshbandi/notation-core-go/blob/core-refactor/signer/signer.go#L139

signer/jws.go
```diff
@@ -0,0 +1,475 @@
...
+		return err
+	}
+
+	if valError := validateSignerInfo(sigInfo); valError != nil {
+		return valError
+	}
+
+	// verify JWT
+	compact := strings.Join([]string{jws.internalEnv.Protected, jws.internalEnv.Payload, jws.internalEnv.Signature}, ".")
+	return verifyJWT(compact, sigInfo.CertificateChain[0].PublicKey)
```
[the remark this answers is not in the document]
priteshbandi: Good catch. Handled it in [jws.go#L418](https://github.com/priteshbandi/notation-core-go/blob/core-refactor/signer/jws.go#L418)

signer/jws.go
```diff
@@ -0,0 +1,475 @@
...
+	}
+
+	return JWSEnvelope{internalEnv: jwsInternal}, nil
+}
+
+func newJWSEnvelope() JWSEnvelope {
+	return JWSEnvelope{internalEnv: newJwsInternalEnvelope()}
+}
+
+func (jws *JWSEnvelope) validateIntegrity() error {
```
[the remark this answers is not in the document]
priteshbandi: The implementation was wrong, `validateIntegrity` method shouldn't create SignerInfo object, it should only parse elements required to validate the integrity of the envelope and nothing beyond that. I've updated the implementation to reflect the same.

But you have a valid point for SignatureEnvelope$Verify method, will update it.

signer/signer.go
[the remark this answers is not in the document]
rgnote: We may want to do It sooner so that we don't accidentally export an internal struct. We just need to move the internal types to their respective files in /internal

signer/signer.go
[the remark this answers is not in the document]
rgnote: Not sure why the callers need it. How will they use the cert?

signer/jws.go
rgnote: We may want to be explicit about this. Can we have a separate method that filters the ExtendedAttributes?

signer/jws.go
```diff
@@ -0,0 +1,480 @@
...
+	if err != nil {
+		return nil, err
+	}
+	var pHeaders map[string]interface{}
+	if err = json.Unmarshal(protected, &pHeaders); err != nil {
+		return nil, err
+	}
+	if err := populateProtectedHeaders(pHeaders, &signInfo); err != nil {
+		return nil, err
+	}
```
qmuntal, line 122: I'm having troubles following the logic in `populateProtectedHeaders`. Why don't we decode the protected header using a `struct` with the appropriate properties and json tags instead of unmarshaling into a map and then manually populating everything? It would be less complex and easier to maintain.

qmuntal (changes requested):
General comment before doing an in-depth review:

Could you take a look at https://github.com/notaryproject/notation-go/pull/62? I've changes the data model of several notation types to fit the new signing spec and moved away from `jwt.Token`, which simplifies the encoding/decoding logic. I think this PR would also benefit from taking that approach, as I guess it is using that many maps just because `jwt.Token.Header` is a map instead of a typed struct.

signer/jws.go
```diff
@@ -0,0 +1,480 @@
...
+	attrs[headerKeyCrit] = crit
+	return attrs
+}
+
+// ***********************************************************************
+// jwsEnvelope-JSON specific code
+// ***********************************************************************
+const (
+	// JWSPayloadContentType describes the media type of the jwsEnvelope envelope.
+	JWSPayloadContentType = "application/vnd.cncf.notary.v2.jws.v1"
```
rgnote, line 293: This shouldn't be exported, correct?

signer/jws.go
```diff
@@ -0,0 +1,480 @@
...
+)
+
+// jwsInternalEnvelope is the final Signature envelope.
+type jwsInternalEnvelope struct {
+	// JWSPayload Base64URL-encoded.
+	Payload string `json:"Payload"`
```
rgnote, line 299: Json fields are lowercase? Ref : https://github.com/notaryproject/notaryproject/blob/main/signature-envelope-jws.md#signature-envelope

signer/jws.go
```diff
@@ -0,0 +1,480 @@
...
+	CertChain [][]byte `json:"x5c"`
+
+	// SigningAgent used for signing
+	SigningAgent string `json:"io.cncf.notary.SigningAgent,omitempty"`
+}
+
+func newJwsInternalEnvelopeFromBytes(b []byte) (*jwsInternalEnvelope, error) {
+	var jws jwsInternalEnvelope
+	if err := json.Unmarshal(b, &jws); err != nil {
+		return &jws, err
```
rgnote, line 327:
```suggestion
		return nil, err
```

signer/jws.go
```diff
@@ -0,0 +1,480 @@
...
+	}
+	return &jws, nil
+}
+
+
+func generateJws(compact string, req SignRequest) (*jwsInternalEnvelope, error) {
+	parts := strings.Split(compact, ".")
+	if len(parts) != 3 {
+		// this should never happen
+		return nil, errors.New("invalid compact serialization")
```
rgnote, line 337: > invalid compact serialization

This can be more descriptive. May be `unexpected error while generating a JWS signature envelope`?

signer/jws.go
```diff
@@ -0,0 +1,480 @@
...
+const (
+	headerKeyExpiry      = "io.cncf.notary.expiry"
+	headerKeySigningTime = "io.cncf.notary.signingTime"
+	headerKeyCrit        = "crit"
+	headerKeyAlg         = "alg"
+	headerKeyCty         = "cty"
+)
+
+// jwsInternalEnvelope represents the jwsEnvelope-JSON envelope
+type jwsEnvelope struct {
```
rgnote, line 32: I wonder why do we need this wrapper around `jwsInternalEnvelope`? We should be able to get the same functionality with just `jwsInternalEnvelope`, right? What am I missing?

signer/jws.go
```diff
@@ -0,0 +1,480 @@
...
+
+const (
+	headerKeyExpiry      = "io.cncf.notary.expiry"
+	headerKeySigningTime = "io.cncf.notary.signingTime"
+	headerKeyCrit        = "crit"
+	headerKeyAlg         = "alg"
+	headerKeyCty         = "cty"
+)
+
+// jwsInternalEnvelope represents the jwsEnvelope-JSON envelope
```
rgnote, line 31:
```suggestion
// jwsEnvelope represents the jwsEnvelope-JSON envelope
```

signer/signer.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	// TODO: perform PayloadContentType value validations
+	if len(req.PayloadContentType) == 0 {
+		return MalformedSignRequestError{msg: "signature content type not present or is empty"}
+	}
+
+	if len(req.CertificateChain) == 0 {
+		return MalformedSignRequestError{msg: "certificate-chain not present or is empty"}
+	}
```
rgnote, line 201: Can we validate the chain is valid here?

signer/jws.go
priteshbandi: Notary v2 allows users to add custom signed attributes and in the case of JWS these custom signed attributes will go into protected headers.
Since we don't know the keys for these custom attributes we cannot use `struct` and that's the reason we need to unmarshall protected headers to a map and then manually populate everything

signer/jws.go
priteshbandi: Its intended to be exported as caller can use it for specifying the contentType during sign operation.

signer/jws.go
```diff
@@ -0,0 +1,480 @@
...
+	if err := json.Unmarshal(req.Payload, &m); err != nil {
+		return nil, err
+	}
+	signingMethod, err := getSigningMethod(leafPublicKey)
+	if err != nil {
+		return nil, err
+	}
+
+	signedAttrs := getSignedAttrs(req, signingMethod)
+	compact, err := signJWT(m, signedAttrs, signingMethod, req.SignatureProvider)
```
qmuntal, line 80: `req.Payload` should already be a properly formatted payload ready to sign, so we could avoid unmarshaling and decoding it into signed signed attribues + claims if we just call `signingMethodForSign{algo: method.Alg(), sigProvider: sigPro}.Sign(req.Payload, "DummyNotUsed")` and then concatenating the resulting signature with the payload.

signer/jws.go
qmuntal: Good point. If that's the case, I would prefer to unmarshall into a typed struct for the known fields and then unmarshall again into a map deleting the known fields.

This way we can get rid of `populateProtectedHeaders`, aka less code to maintain, and we only pay an, and probably unnoticeable, performance penalty.

signer/signer.go
```diff
@@ -0,0 +1,257 @@
...
+	// getSignerInfo returns the information stored in the Signature envelope and doesn't perform integrity verification.
+	getSignerInfo() (*SignerInfo, error)
+	// signPayload created Signature envelope.
+	signPayload(SignRequest) ([]byte, error)
+}
+
+// Verify performs integrity validation and validates that cert-chain stored in Signature leads to a given set of the trusted root.
+// Returns the certificate that is used for establishing trust i.e. authenticity of the signature, SignerInfo object
+// containing the information about the signature.
+func (s *SignatureEnvelope) Verify(trustedCerts []*x509.Certificate) (*x509.Certificate, *SignerInfo, error) {
```
qmuntal, line 89: Why do we return `SignerInfo` here? It complicates the function signature for a use-case that is already covered by calling `GetSignerInfo()`, and it adds overhead to the function implementation.

signer/signer.go
```diff
@@ -0,0 +1,257 @@
...
+	if len(trustedCerts) == 0 {
+		return nil, nil, MalformedArgumentError{param: "trustedCerts"}
+	}
+
+	integrityError := s.internalEnvelope.validateIntegrity()
+	if integrityError != nil {
+		return nil, nil, integrityError
+	}
+
+	singerInfo, singerInfoErr := s.GetSignerInfo()
```
qmuntal, line 103: If we end up not returning the `SignerInfo`, as I proposed in a previous comment, there is no need to decode the whole envelope, just the unprotected header containing the certificate chain.

qmuntal (changes requested):

signer/signer.go
```diff
@@ -0,0 +1,257 @@
...
+	ECDSA_SHA_384      SignatureAlgorithm = "ECDSA_SHA_384"
+	ECDSA_SHA_512      SignatureAlgorithm = "ECDSA_SHA_512"
+)
+
+// SignerInfo represents a parsed signature envelope that is agnostic to signature envelope format.
+type SignerInfo struct {
+	Payload            []byte
+	PayloadContentType string
+	SignedAttributes   SignedAttributes
+	UnsignedAttributes UnsignedAttributes
```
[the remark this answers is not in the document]
priteshbandi: CertificateChain and TimestampSignature are intentionally kept out as they are not attributes of the signature but rather an important part of the signature(they complete or countersign the signature). Also, `SignerInfo` is not tied to JWS, rather it's an abstract type independent of signature envelope format.

signer/signer.go
priteshbandi: Good catch...will do.

signer/jws.go
priteshbandi: Updated code to unmarshall into a typed struct for the known fields and then unmarshall again into a map deleting the known fields.

signer/signer.go
priteshbandi: decoding of envelope is required to validate that it contains required signed attributes and there values are valid.

signer/jws.go
```diff
@@ -0,0 +1,480 @@
...
+}
+
+func populateString(data map[string]interface{}, s string, holder *string) error {
+	if val, ok := data[s]; ok {
+		valString, ok := val.(string)
+		if !ok {
+			return MalformedSignatureError{msg: fmt.Sprintf("failed to parse %q in protected headers", s)}
+		}
+		*holder = valString
+		delete(data, s)
```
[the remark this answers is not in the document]
priteshbandi: Updated logic to use well defined types.

signer/jws.go
```diff
@@ -0,0 +1,480 @@
...
+		return ECDSA_SHA_384, nil
+	case "ES512":
+		return ECDSA_SHA_512, nil
+	}
+
+	return RSASSA_PSS_SHA_512, SignatureAlgoNotSupportedError{alg: alg}
+}
+
+// ***********************************************************************
+// JWT specific code
```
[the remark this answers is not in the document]
priteshbandi: yeah why not.

signer/jws.go
priteshbandi: Move to well defined struct...code path is no longer relevant.

qmuntal (changes requested):

signer/jws.go
```diff
@@ -0,0 +1,396 @@
...
+func newJWSEnvelopeFromBytes(envelopeBytes []byte) (*jwsEnvelope, error) {
+	jwsInternal, err := newJwsInternalEnvelopeFromBytes(envelopeBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	return &jwsEnvelope{internalEnv: jwsInternal}, nil
+}
+
+func newJWSEnvelope() *jwsEnvelope {
```
qmuntal, line 41: Declaring a constructor function that just wraps an object declaration is not idiomatic. Just use `new(jwsEnvelope)`.

qmuntal (changes requested):

internal/testhelper/certificatetest.go
```diff
@@ -0,0 +1,147 @@
...
+// GetECLeafCertificate returns leaf certificate signed using EC algorithm
+func GetECLeafCertificate() ECCertTuple {
+	return ecdsaLeaf
+}
+
+// GetUnsupportedCertificate returns certificate signed using RSA algorithm with key size of 1024 bits
+// which is not supported by notary.
+func GetUnsupportedCertificate() CertTuple {
+	return unsupported
+}
```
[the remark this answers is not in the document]
priteshbandi: Having global gives less control than the function, because anyone can read or change the global variable at any time. Also, having function will allow us to change variable name/representation internally without affecting callers.

signer/jws.go
priteshbandi: This will allow us to replace the `jwsInternalEnvelope` with a JWS-JSON library(If we ever find) without having to much refactoring.

Also, doing it this way provides better code organization.

rgnote: Replacing `jwsInternalEnvelope` with JWS-Json library is same effort with or without `jwsEnvelope`. I feel `jwsEnvelope` is an unnecessary abstraction with no real value. Not sure about better code organization but it is making the code hard to read.

signer/signer.go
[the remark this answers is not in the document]
rgnote: Fair enough

signer/jws_test.go
```diff
@@ -0,0 +1,148 @@
...
+	"crypto/x509"
+	"errors"
+	"github.com/notaryproject/notation-core-go/internal/testhelper"
+	"testing"
+)
+
+func TestNewJWSEnvelopeFromBytes(t *testing.T) {
+	t.Run("newJWSEnvelopeFromBytes", func(t *testing.T) {
+		if _, err := newJWSEnvelopeFromBytes([]byte(TestValidSig)); err != nil {
+			t.Errorf("Error found")
```
rgnote, line 13: Print err? Also, good idea to print the test case name.

signer/jws_test.go
```diff
@@ -0,0 +1,148 @@
...
+		err := env.validateIntegrity()
+		if err != nil {
+			t.Errorf("validateIntegrity(). Error = %s", err)
+		}
+	})
+
+	t.Run("with invalid base64 bytes sig envelope returns error", func(t *testing.T) {
+		env, _ := newJWSEnvelopeFromBytes([]byte("{\"Payload\":\"Hi!\",\"Protected\":\"Hi\",\"Header\":{},\"Signature\":\"Hi!\"}"))
+		if err := env.validateIntegrity(); err == nil {
+			t.Errorf("Expected error but not found")
```
rgnote, line 44: Can we also test the type of the expected err?

signer/jws_test.go
```diff
@@ -0,0 +1,148 @@
...
+	})
+
+	t.Run("newJWSEnvelopeFromBytes Error", func(t *testing.T) {
+		if _, err := newJWSEnvelopeFromBytes([]byte("Malformed")); err == nil {
+			t.Errorf("Expected error but not found")
+		}
+	})
+}
+
+func TestValidateIntegrity(t *testing.T) {
```
rgnote, line 24: Can we add a test case with a signature and an unrelated leaf cert?

signer/jws_test.go
```diff
@@ -0,0 +1,148 @@
...
+		env, _ := newJWSEnvelopeFromBytes([]byte("{\"Payload\":\"Hi!\",\"Protected\":\"Hi\",\"Header\":{},\"Signature\":\"Hi!\"}"))
+		if err := env.validateIntegrity(); err == nil {
+			t.Errorf("Expected error but not found")
+		}
+	})
+
+	t.Run("with incomplete sig envelope returns error", func(t *testing.T) {
+		env, _ := newJWSEnvelopeFromBytes([]byte("{\"Payload\":\"eyJhbGciOiJIUzI1NiJ9\",\"Protected\":\"eyJhbGciOiJQUzI1NiIsImNyaXQiOlsiaW8uY25jZi5ub3Rhcnkuc2lnbmluZ1RpbWUiXSwiaW8uY25jZi5ub3Rhcnkuc2luaW5nVGltZSI6IjIwMDYtMDEtMDJUMTU6MDQ6MDVaIn0\",\"Header\":{},\"Signature\":\"YjGj\"}"))
+		if err := env.validateIntegrity(); !(err != nil && errors.As(err, new(MalformedSignatureError))) {
+			t.Errorf("Expected MalformedSignatureError but but found %q", err)
```
rgnote, line 51: nit: double "but"

signer/jws_test.go
```diff
@@ -0,0 +1,148 @@
...
+		if !(err != nil && errors.As(err, new(SignatureNotFoundError))) {
+			t.Errorf("Expected SignatureNotFoundError but found %q", err)
+		}
+	})
+
+	t.Run("with newJWSEnvelope after sign sign works", func(t *testing.T) {
+		env := jwsEnvelope{}
+		_, err := env.getSignerInfo()
+		env.signPayload(getSignRequest())
+		if !(err != nil && errors.As(err, new(SignatureNotFoundError))) {
```
rgnote, line 69: We should get the SignerInfo after signing the payload, correct?

signer/jws_test.go
```diff
@@ -0,0 +1,148 @@
...
+func TestGetSignerInfo(t *testing.T) {
+	t.Run("with newJWSEnvelope before sign returns error", func(t *testing.T) {
+		env := jwsEnvelope{}
+		_, err := env.getSignerInfo()
+		if !(err != nil && errors.As(err, new(SignatureNotFoundError))) {
+			t.Errorf("Expected SignatureNotFoundError but found %q", err)
+		}
+	})
+
+	t.Run("with newJWSEnvelope after sign sign works", func(t *testing.T) {
```
rgnote, line 65:
```suggestion
	t.Run("with newJWSEnvelope after sign works", func(t *testing.T) {
```

signer/jws_test.go
```diff
@@ -0,0 +1,148 @@
...
+		env := jwsEnvelope{}
+		_, err := env.getSignerInfo()
+		env.signPayload(getSignRequest())
+		if !(err != nil && errors.As(err, new(SignatureNotFoundError))) {
+			t.Errorf("getSignerInfo(). Error = %s", err)
```
rgnote, line 70: You can use `t.Name()` in the error messages instead of hard coding the method names you are testing. That way, it is easier for devs to find a failing unit test.

signer/jws_test.go
```diff
@@ -0,0 +1,148 @@
...
+		_, err := env.getSignerInfo()
+		if err != nil {
+			t.Errorf("getSignerInfo(). Error = %s", err)
+		}
+	})
+
+	t.Run("with invalid base64 bytes sig envelope returns error", func(t *testing.T) {
+		env, _ := newJWSEnvelopeFromBytes([]byte("{\"Payload\":\"Hi!\",\"Protected\":\"Hi\",\"Header\":{},\"Signature\":\"Hi!\"}"))
+		if _, err := env.getSignerInfo(); err == nil {
+			t.Errorf("Expected error but not found")
```
rgnote, line 85: Print test case name

signer/jws_test.go
priteshbandi: yep, done

signer/jws_test.go
priteshbandi: yep updated.

signer/signer.go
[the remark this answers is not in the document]
priteshbandi: As per conversation over call, we will move to internal package after introducing new sig format(cose?) if feel that the package bloated or if we want to share multiple method across multiple packages.

signer/signer_test.go
```diff
@@ -0,0 +1,398 @@
...
+		"		\"io.cncf.notary.SigningAgent\":\"NotationUnitTest/1.0.0\"}," +
+		"\"signature\":\"J9iQDfXM1GYzIazRH36DUgjeBSp1YIv5gqb0evyrp46mRNdsvGxzBvqVM3K1ZYW530wryweL51oVTbXMEh2PWchQZ7g33Be5lgcl82il7rR5D5tpsiSZ7oZsD4LP-Swv6MoYlKW4hKXWTCY9cWLzJhHkGZPiLsyrWUqdBq_0M8BTyx42_MUmAbYrFVKRjZy8PKsFDAaBcZVbdyZWRqVJy4Lfw8n4P0Ry7bDWRkqhI2rXH4o68eSkNF3KGWzQWXTp6uZb7o5HKc3dn3uoNidKvP3kZaM-XfM9Hd9Cw1MxLwvu1Qdjo6MCOatMKxc02cI7LAA6AsRcYfR-vGkVW3bJP9L29GJ-Dufv7dWcC_xCEG7p6lSYcF86haY_iTwSv_IQoKXQrMnwL8yZpbshJBrdjOzojdZBsJ4_Pu7KdNsnTpR-UvnFdIUrPvYek5WwI8jLz9hTVsSzF0aWCnCf7t8sAaUf90CC04kwGP2jnvZKlNcTQpZ56Zl-n43Z6KkC62do\"}\n"
+)
+
+var (
+	TestTemperedSig = strings.Replace(TestValidSig, "0fX0", "1fX0=", 1)
```
rgnote, line 33:
```suggestion
	TestTamperedSig = strings.Replace(TestValidSig, "0fX0", "1fX0=", 1)
```

rgnote (approved):
LGTM

signer/signer.go
```diff
@@ -0,0 +1,261 @@
...
+	Critical bool
+	Value    interface{}
+}
+
+// SignatureProvider is used to sign bytes generated after creating Signature envelope.
+type SignatureProvider interface {
+	Sign([]byte) ([]byte, error)
+}
+
+type SignatureEnvelope struct {
```
shizhMSFT, line 71: Why not make `SignatureEnvelope` an interface lile
```golang
type SignatureEnvelope interface {
    Sign(req SignRequest) ([]byte, error)
    Verify(trustedCerts []*x509.Certificate) (*x509.Certificate, *SignerInfo, error)
    GetSignerInfo() (*SignerInfo, error)
}
```

Then we can have stuffs simplified like

```go
// NewSignatureEnvelope is used for signature generation workflow
func NewSignatureEnvelope(envelopeMediaType SignatureMediaType) (SignatureEnvelope, error) {
	switch envelopeMediaType {
	case MediaTypeJWSJSON:
		return &jwsEnvelope{}, nil
	case MediaTypeCOSE:
		return &coseEnvelope{}, nil
	}
	return nil, &ErrorUnsupportedSignatureFormat{mediaType: string(envelopeMediaType)}
}
```

internal/testhelper/certificatetest.go
```diff
@@ -0,0 +1,147 @@
...
+
+func setupCertificates() {
+	rsaRoot = getCertTuple("Notation Test Root", nil)
+	rsaLeaf = getCertTuple("Notation Test Leaf Cert", &rsaRoot)
+	ecdsaRoot = getECCertTuple("Notation Test Root2", nil)
+	ecdsaLeaf = getECCertTuple("Notation Test Leaf Cert", &ecdsaRoot)
+
+	k, _ := rsa.GenerateKey(rand.Reader, 1024)
```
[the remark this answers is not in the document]
priteshbandi: This is intentional for the unit test. We want to test a scenario where if the user tries to sign with the key size of 1024 notation should return an error

signer/jwt.go
```diff
@@ -0,0 +1,113 @@
...
+	compact, err := token.SignedString("DummyNotUsed")
+	if err != nil {
+		return "", err
+	}
+	return compact, nil
+}
+
+// verifyJWT verifies the JWT token against the specified verification key
+func verifyJWT(tokenString string, key crypto.PublicKey) error {
+	signingMethod, _ := getSigningMethod(key)
```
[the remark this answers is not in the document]
priteshbandi: Good catch. fixed.

signer/errors.go
```diff
@@ -0,0 +1,103 @@
+package signer
+
+import "fmt"
+
+// SignatureIntegrityError is used when the Signature associated is no longer valid.
+type SignatureIntegrityError struct {
```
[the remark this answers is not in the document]
priteshbandi: Looks like we are following convention where Error is in suffix.
https://github.com/notaryproject/notation-go/blob/407537596ed5c9149b4a7a9358bc73042dbb95e9/plugin/errors.go#L41-L45
https://github.com/notaryproject/notation-go/blob/9b9dcd31a2bae1172e457aed50086e6ad0a9f11d/internal/crypto/cms/errors.go#L19-L62

Also there are multiple references in golang code with`Error` as suffix.

signer/signer.go
priteshbandi: So basically we want to implement common validations(like input validations, cert chain validations, trust anchor related code, timestamping related code) in SignatureEnvelope's sign and verify method and let specific implementation only do the work of encoding/decoding/integrity validation signature format.
With interface aboce things are not possible. Although there can be workaround by creating helper methods but than we need to make sure each implementation is calling them.

signer/jwt.go
```diff
@@ -0,0 +1,109 @@
...
+	"crypto/ecdsa"
+	"crypto/rsa"
+	"encoding/base64"
+	"fmt"
+
+	"github.com/golang-jwt/jwt/v4"
+)
+
+// signJWT signs the given payload and headers using the given signing method and signature provider
+func signJWT(payload map[string]interface{}, headers map[string]interface{}, method jwt.SigningMethod, sigPro SignatureProvider) (string, error) {
```
shizhMSFT, line 14: nit:
```suggestion
func signJWT(payload, headers map[string]interface{}, method jwt.SigningMethod, sigPro SignatureProvider) (string, error) {
```

signer/jwt.go
```diff
@@ -0,0 +1,109 @@
...
+
+	token.Method = signingMethodForSign{algo: method.Alg(), sigProvider: sigPro}
+
+	// We are using custom signing method called `signingMethodForSign` which already has signature provider
+	// thus we don't need to pass signing key as input.
+	compact, err := token.SignedString(nil)
+	if err != nil {
+		return "", err
+	}
+	return compact, nil
```
shizhMSFT, line 29: nit:
```suggestion
	return token.SignedString(nil)
```

signer/jwt.go
```diff
@@ -0,0 +1,109 @@
...
+		// override default signing method with key-specific method
+		t.Method = signingMethod
+		return key, nil
+	}); err != nil {
+		return SignatureIntegrityError{err: err}
+	}
+	return nil
+}
+
+// signingMethodForSign It's only used during signature generation operation. It's required by JWT library we are using
```
shizhMSFT, line 60: nit:
```suggestion
// signingMethodForSign is only used during signature generation operation. It's required by JWT library we are using
```

signer/signer.go
```diff
@@ -0,0 +1,262 @@
...
+
+import (
+	"crypto/x509"
+	"fmt"
+	"time"
+
+	nx509 "github.com/notaryproject/notation-core-go/x509"
+)
+
+type SignatureMediaType string
```
shizhMSFT, line 11: Missing documentation for `SignatureMediaType` and other exported structs.

signer/signer.go
```diff
@@ -0,0 +1,262 @@
...
+	CertificateChain   []*x509.Certificate
+	Signature          []byte
+	TimestampSignature []byte
+}
+
+// SignedAttributes represents signed metadata in the Signature envelope
+type SignedAttributes struct {
+	SigningTime        time.Time
+	Expiry             time.Time
+	ExtendedAttributes []Attribute
```
shizhMSFT, line 41: What if we have two attributes with the same key?

signer/signer.go
shizhMSFT: Then we can still have

```go
// NewSignatureEnvelope is used for signature generation workflow
func NewSignatureEnvelope(envelopeMediaType SignatureMediaType) (SignatureEnvelope, error) {
	switch envelopeMediaType {
	case MediaTypeJWSJSON:
		return &someGenericValidation{&jwsEnvelope{}}, nil
	case MediaTypeCOSE:
		return &someGenericValidation{&coseEnvelope{}}, nil
	}
	return nil, &UnsupportedSignatureFormatError{mediaType: string(envelopeMediaType)}
}
```

That still works. With interfaces, it is easier to decouple code and do unit testing.

signer/signer.go
```diff
@@ -0,0 +1,262 @@
...
+		return nil, err
+	}
+
+	sig, err := s.internalEnvelope.signPayload(req)
+	if err != nil {
+		return nil, err
+	}
+
+	s.rawSignatureEnvelope = sig
+	return sig, err
```
shizhMSFT, line 135: nit: 
```suggestion
	return sig, nil
```

signer/signer.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	// TODO: perform PayloadContentType value validations
+	if len(info.PayloadContentType) == 0 {
+		return MalformedSignatureError{msg: "signature content type not present or is empty"}
+	}
+
+	signTime := info.SignedAttributes.SigningTime
+	if signTime.IsZero() {
+		return MalformedSignatureError{msg: "singing-time not present"}
```
shizhMSFT, line 168:
```suggestion
		return MalformedSignatureError{msg: "signing-time not present"}
```

signer/signer.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	expTime := info.SignedAttributes.Expiry
+	if !expTime.IsZero() {
+		if expTime.Before(signTime) || expTime.Equal(signTime) {
+			return MalformedSignatureError{msg: "expiry cannot be equal or before the signing time"}
+		}
+	}
```
shizhMSFT, line 176:
```suggestion
	if !expTime.IsZero() && !expTime.After(signTime) {
		return MalformedSignatureError{msg: "expiry cannot be equal or before the signing time"}
	}
```

signer/signer.go
```diff
@@ -0,0 +1,262 @@
...
+
+	if len(info.Signature) == 0 {
+		return MalformedSignatureError{msg: "signature not present or is empty"}
+	}
+
+	return nil
+}
+
+// validateSignRequest performs basic set of validations on SignRequest struct.
+func validateSignRequest(req SignRequest) error {
```
shizhMSFT, line 190: It seems that there are duplicated codes in `validateSignerInfo` and `validateSignRequest`. Can we refactor them?

signer/signer.go
```diff
@@ -0,0 +1,262 @@
...
+
+	singerInfo, singerInfoErr := s.GetSignerInfo()
+	if singerInfoErr != nil {
+		return nil, nil, singerInfoErr
+	}
+
+	certChain := singerInfo.CertificateChain
+	err := nx509.ValidateCertChain(certChain)
+	if err != nil {
+		return nil, nil, singerInfoErr
```
rgnote, line 112:
```suggestion
		return nil, nil, err
```

signer/signer.go
```diff
@@ -0,0 +1,262 @@
...
+	// getSignerInfo returns the information stored in the Signature envelope and doesn't perform integrity verification.
+	getSignerInfo() (*SignerInfo, error)
+	// signPayload created Signature envelope.
+	signPayload(SignRequest) ([]byte, error)
+}
+
+// Verify performs integrity validation and validates that cert-chain stored in Signature leads to a given set of the trusted root.
+// Returns the certificate that is used for establishing trust i.e. authenticity of the signature, SignerInfo object
+// containing the information about the signature.
+func (s *SignatureEnvelope) Verify(trustedCerts []*x509.Certificate) (*x509.Certificate, *SignerInfo, error) {
```
rgnote, line 90: Had an offline discussion to move "authenticity" part of this function into it's own method for flexibility.

signer/signer.go
priteshbandi: If we have two attributes with same key we return two attributes. Its JWS that restricts protected headers to not have duplicate keys but other formats like [PKCS7](https://datatracker.ietf.org/doc/html/rfc2315#section-9.2) allows duplicate keys in attributes.

signer/signer.go
priteshbandi: We also want equality check. i.e. signing time and expiry time should not be equal

signer/signer.go
priteshbandi: Done

signer/signer.go
priteshbandi: Added

shizhMSFT (approved):
LGTM

internal/testhelper/certificatetest.go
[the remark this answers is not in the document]
gokarnm: Can you add a comment that this is flagged by static analysis, and that it's for test purpose. We probably need to figure out how to suppress warning for this specific case.

signer/signer.go
```diff
@@ -0,0 +1,251 @@
...
+	validateIntegrity() error
+	// getSignerInfo returns the information stored in the Signature envelope and doesn't perform integrity verification.
+	getSignerInfo() (*SignerInfo, error)
+	// signPayload created Signature envelope.
+	signPayload(SignRequest) ([]byte, error)
+}
+
+// Verify performs integrity and other signature specification related validations
+// Returns the SignerInfo object containing the information about the signature.
+func (s *SignatureEnvelope) Verify() (*SignerInfo, error) {
```
gokarnm, line 92: Typo `singerInfo`, can be addressed in subsequent PR.

gokarnm (approved):

gokarnm:
LGTM!

qmuntal (approved):

signer/signer.go
```diff
@@ -0,0 +1,251 @@
...
+	SigningAgent string
+}
+
+// SignRequest is used to generate Signature.
+type SignRequest struct {
+	Payload             []byte
+	PayloadContentType  string
+	CertificateChain    []*x509.Certificate
+	SignatureProvider   SignatureProvider
+	SigningTime         time.Time
```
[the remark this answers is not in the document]
priteshbandi: Its an library that would be called by product which should be able to specify the signing time.

PR #8, opened 2022-06-17
Setup Github workflow

Wwwsylvia:
Signed-off-by: Lixia (Sylvia) Lei <lixia_lei@outlook.com>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-name: build
+name: Build
 
 on:
   push:
@@ -8,23 +8,4 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        go-version: [1.17, 1.18]
-      fail-fast: true
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v3
-      - name: Set up Go ${{ matrix.go-version }} environment
-        uses: actions/setup-go@v3
-        with:
-          go-version: ${{ matrix.go-version }}
-          check-latest: true
-      - name: Run unit tests
-        run: make test
-      - name: Upload coverage report
-        uses: actions/upload-artifact@master
-        with:
-          name: notation-coverage-report-${{ github.sha }}
-          path: .cover/
+    uses: ./.github/workflows/reusable-build.yml
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,30 @@
+name: build
+
+on:
+  push:
+    branches: main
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Set up Go ${{ matrix.go-version }} environment
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ matrix.go-version }}
+          check-latest: true
+      - name: Run unit tests
+        run: make test
```
gokarnm, line 25: I'm curious why the steps across [notation](https://github.com/notaryproject/notation/blob/main/.github/workflows/build.yml#L23-L34), [notation-go](https://github.com/notaryproject/notation-go/pull/60/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R25-R33), and notation-core-go are slightly different, can they all be the same set of steps. Can they be [reused workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows#reusable-workflows-and-starter-workflows) defined in a single place?

.github/workflows/build.yml
```diff
@@ -0,0 +1,30 @@
...
+
+on:
+  push:
+    branches: main
+  pull_request:
+    branches: main
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
gokarnm, line 11: We should update notation [build.yml](https://github.com/notaryproject/notation/blob/main/.github/workflows/build.yml#L11) to use the same.

rgnote: Can we also build/test on Windows and Mac?

Wwwsylvia: Make sense. But we may want to do this in a seperate PR, as currently we use make test to generate coverage report, which will not work on Windows.

.github/workflows/build.yml
Wwwsylvia: Good idea!

.github/workflows/build.yml
Wwwsylvia: @gokarnm Updated notation workflow. See https://github.com/notaryproject/notation/pull/212/files

.github/workflows/build.yml
Wwwsylvia: Defined a reusable workflow.

gokarnm (approved):
LGTM!

shizhMSFT (approved):
LGTM

PR #11, opened 2022-06-21
Remove opencontainer dependency for timestamping code

priteshbandi:
* timestamping shouldn't depend on opencontainers.
* Also, rephrased some of the error message for certificate chain validation.

### DIFF
--- a/timestamp/http_test.go
+++ b/timestamp/http_test.go
@@ -75,9 +75,9 @@ func TestHTTPTimestampGranted(t *testing.T) {
 		t.Fatalf("NewHTTPTimestamper() error = %v", err)
 	}
 	message := []byte("notation")
-	req, err := NewRequestWithContent(message, crypto.SHA256)
+	req, err := NewRequestFromContent(message, crypto.SHA256)
 	if err != nil {
-		t.Fatalf("NewRequestWithContent() error = %v", err)
+		t.Fatalf("NewRequestFromContent() error = %v", err)
 	}
 	ctx := context.Background()
 	resp, err := tsa.Timestamp(ctx, req)
@@ -128,7 +128,7 @@ func TestHTTPTimestampGranted(t *testing.T) {
 	if err != nil {
 		t.Fatal("SignedToken.Info() error =", err)
 	}
-	if err := info.VerifyWithContent(message); err != nil {
+	if err := info.VerifyContent(message); err != nil {
 		t.Errorf("TSTInfo.Verify() error = %v", err)
 	}
 	timestamp, accuracy := info.Timestamp()
@@ -173,9 +173,9 @@ func TestHTTPTimestampRejection(t *testing.T) {
 		t.Fatalf("NewHTTPTimestamper() error = %v", err)
 	}
 	message := []byte("notation")
-	req, err := NewRequestWithContent(message, crypto.SHA256)
+	req, err := NewRequestFromContent(message, crypto.SHA256)
 	if err != nil {
-		t.Fatalf("NewRequestWithContent() error = %v", err)
+		t.Fatalf("NewRequestFromContent() error = %v", err)
 	}
 	ctx := context.Background()
 	resp, err := tsa.Timestamp(ctx, req)
@@ -216,9 +216,9 @@ func TestHTTPTimestampBadEndpoint(t *testing.T) {
 		t.Fatalf("NewHTTPTimestamper() error = %v", err)
 	}
 	message := []byte("notation")
-	req, err := NewRequestWithContent(message, crypto.SHA256)
+	req, err := NewRequestFromContent(message, crypto.SHA256)
 	if err != nil {
-		t.Fatalf("NewRequestWithContent() error = %v", err)
+		t.Fatalf("NewRequestFromContent() error = %v", err)
 	}
 	ctx := context.Background()
 	_, err = tsa.Timestamp(ctx, req)
@@ -240,9 +240,9 @@ func TestHTTPTimestampEndpointNotFound(t *testing.T) {
 		t.Fatalf("NewHTTPTimestamper() error = %v", err)
 	}
 	message := []byte("notation")
-	req, err := NewRequestWithContent(message, crypto.SHA256)
+	req, err := NewRequestFromContent(message, crypto.SHA256)
 	if err != nil {
-		t.Fatalf("NewRequestWithContent() error = %v", err)
+		t.Fatalf("NewRequestFromContent() error = %v", err)
 	}
 	ctx := context.Background()
 	_, err = tsa.Timestamp(ctx, req)
--- a/timestamp/request.go
+++ b/timestamp/request.go
@@ -61,8 +61,8 @@ func NewRequest(digest []byte, alg crypto.Hash) (*Request, error) {
 	}, nil
 }
 
-// NewRequestWithContent creates a request based on the given data and hash algorithm.
-func NewRequestWithContent(content []byte, alg crypto.Hash) (*Request, error) {
+// NewRequestFromContent creates a request based on the given data and hash algorithm.
+func NewRequestFromContent(content []byte, alg crypto.Hash) (*Request, error) {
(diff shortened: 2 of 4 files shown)


jondonas (approved):

rgnote (approved):
LGTM

timestamp/request.go
```diff
@@ -85,15 +88,34 @@ func (r *Request) UnmarshalBinary(data []byte) error {
...
+func validate(digest []byte, alg crypto.Hash) error {
+	l := len(digest)
+	var validContent bool
+	switch alg {
+	case crypto.SHA256:
+		validContent = l == 256/8
+	case crypto.SHA384:
+		validContent =  l == 384/8
+	case crypto.SHA512:
+		validContent = l == 384/8
```
gokarnm, line 113: We are missing unit tests.

```suggestion
	case crypto.SHA512:
		validContent = l == 512/8
```

timestamp/http_test.go
```diff
@@ -75,7 +75,7 @@ func TestHTTPTimestampGranted(t *testing.T) {
 		t.Fatalf("NewHTTPTimestamper() error = %v", err)
 	}
 	message := []byte("notation")
-	req, err := NewRequestFromBytes(message)
+	req, err := NewRequestWithData(message, crypto.SHA256)
 	if err != nil {
 		t.Fatalf("NewRequestFromString() error = %v", err)
```
gokarnm, line 80: Method name on line 80 does not match line 78. Same for multiple occurrences in this file.

timestamp/request.go
```diff
@@ -37,36 +38,38 @@ type Request struct {
...
 		CertReq: true,
 	}, nil
 }
 
-// NewRequestFromBytes creates a request based on the given byte slice.
-func NewRequestFromBytes(content []byte) (*Request, error) {
-	return NewRequest(digest.FromBytes(content))
-}
+// NewRequestWithData creates a request based on the given data and hash algorithm.
+func NewRequestWithData(content []byte, alg crypto.Hash) (*Request, error) {
```
gokarnm, line 66: Suggestion - NewRequestWithContent ?

timestamp/http_test.go
gokarnm: NewRequestFromString is removed but multiple error messages reference it.

timestamp/request.go
```diff
@@ -37,36 +38,38 @@ type Request struct {
 	Extensions     []pkix.Extension      `asn1:"optional,tag:0"`
 }
 
-// NewRequest creates a request based on the given digest.
-func NewRequest(contentDigest digest.Digest) (*Request, error) {
-	hashAlgorithm, err := getOIDFromDigestAlgorithm(contentDigest.Algorithm())
+// NewRequest creates a request based on the given digest and hash algorithm.
+func NewRequest(digest []byte, alg crypto.Hash) (*Request, error) {
```
gokarnm, line 42: Is there a reason we need both NewRequest/NewRequestWithData , Verify/VerifyWithData ?

gokarnm (changes requested):

timestamp/errors.go
```diff
@@ -0,0 +1,10 @@
+package timestamp
+
+// MalformedRequestError is used when timestamping request is malformed.
+type MalformedRequestError struct {
```
shizhMSFT, line 4: It is better to name it `ErrorMalformedRequest`. It's a convention to put `Error` as a prefix.

timestamp/request.go
```diff
@@ -1,14 +1,15 @@
 package timestamp
 
 import (
+	"crypto"
 	"crypto/x509/pkix"
 	"encoding/asn1"
-	"encoding/hex"
 	"errors"
+	"fmt"
+	"github.com/notaryproject/notation-core-go/internal/crypto/hashutil"
```
shizhMSFT, line 9: This line should not be with the block of import for built-in packages.

timestamp/token.go
```diff
@@ -7,11 +7,11 @@ import (
 	"encoding/asn1"
 	"errors"
 	"fmt"
+	"github.com/notaryproject/notation-core-go/internal/crypto/hashutil"
```
shizhMSFT, line 10: same here

timestamp/token.go
```diff
@@ -106,8 +106,8 @@ type TSTInfo struct {
 	Extensions     []pkix.Extension `asn1:"optional,tag:1"`
 }
 
-// Verify verifies the message against the timestamp token information.
-func (tst *TSTInfo) Verify(message []byte) error {
+// VerifyWithData verifies the message against the timestamp token information.
+func (tst *TSTInfo) VerifyWithData(message []byte) error {
```
shizhMSFT, line 110: Will `VerifyMessage` be a better name?

timestamp/request.go
```diff
@@ -85,15 +88,34 @@ func (r *Request) UnmarshalBinary(data []byte) error {
...
+	switch alg {
+	case crypto.SHA256:
+		validContent = l == 256/8
+	case crypto.SHA384:
+		validContent =  l == 384/8
+	case crypto.SHA512:
+		validContent = l == 384/8
+	default:
+		return MalformedRequestError{msg: fmt.Sprintf("unsupported hashing algorithm: %s", alg)}
+	}
```
shizhMSFT, line 116: It is better to have `alg.Size()` than hard coded length. 

See https://pkg.go.dev/crypto#Hash.Size

priteshbandi: Done

timestamp/token.go
priteshbandi: Renamed it to VerifyWithContent.

timestamp/request.go
```diff
@@ -37,36 +38,37 @@ type Request struct {
...
 		CertReq: true,
 	}, nil
 }
 
-// NewRequestFromBytes creates a request based on the given byte slice.
-func NewRequestFromBytes(content []byte) (*Request, error) {
-	return NewRequest(digest.FromBytes(content))
-}
+// NewRequestWithContent creates a request based on the given data and hash algorithm.
+func NewRequestWithContent(content []byte, alg crypto.Hash) (*Request, error) {
```
shizhMSFT, line 65: nit:
```suggestion
func NewRequestFromContent(content []byte, alg crypto.Hash) (*Request, error) {
```

timestamp/token.go
```diff
@@ -106,8 +106,8 @@ type TSTInfo struct {
 	Extensions     []pkix.Extension `asn1:"optional,tag:1"`
 }
 
-// Verify verifies the message against the timestamp token information.
-func (tst *TSTInfo) Verify(message []byte) error {
+// VerifyWithContent verifies the message against the timestamp token information.
+func (tst *TSTInfo) VerifyWithContent(message []byte) error {
```
shizhMSFT, line 110: nit:
```suggestion
func (tst *TSTInfo) VerifyContent(message []byte) error {
```

timestamp/request.go
```diff
@@ -85,15 +87,34 @@ func (r *Request) UnmarshalBinary(data []byte) error {
...
+	case crypto.SHA512:
+		validContent = l == crypto.SHA512.Size()
+	default:
+		return MalformedRequestError{msg: fmt.Sprintf("unsupported hashing algorithm: %s", alg)}
+	}
+	if !validContent {
+		return MalformedRequestError{msg: fmt.Sprintf("digest is of incorrect size: %d", l)}
+	}
+	return nil
+}
```
shizhMSFT, line 120:
```suggestion
func validate(digest []byte, alg crypto.Hash) error {
	switch alg {
	case crypto.SHA256, crypto.SHA384, crypto.SHA512:
		// no-op
	default:
		return MalformedRequestError{msg: fmt.Sprintf("unsupported hashing algorithm: %s", alg)}
	}
	if len(digest) != alg.Size() {
		return MalformedRequestError{msg: fmt.Sprintf("digest is of incorrect size: %d", l)}
	}
	return nil
}
```

gokarnm (approved):

PR #15, opened 2022-07-07
Update SignatureProvider to also return certificate chain and other minor changes.

priteshbandi:
### Change log
* Update SignatureProvider to also return certificate chain
* Added LocalSignatureProvider to facilitate signing using local key and certificates.
* Moved testhelpers out of internal

Signed-off-by: Pritesh Bandi <pritesb@amazon.com>

PR #17, opened 2022-07-12
Implement signing scheme.

priteshbandi:
### Change Log
1. Implement signature scheme. Specs in https://github.com/notaryproject/notaryproject/pull/165
2. address comment of PR #15

### Testing
Unit tests.

### Signature example
Minimal Signature: https://tinyurl.com/878aj5vd
Loaded Signature: https://tinyurl.com/2p8397vb


signer/signer.go
```diff
@@ -34,13 +38,16 @@ type UnsignedAttributes struct {
...
+	Payload                      []byte
+	PayloadContentType           PayloadContentType
+	SignatureProvider            SignatureProvider
+	SigningTime                  time.Time
+	Expiry                       time.Time
+	ExtendedSignedAttrs          []Attribute
+	SigningAgent                 string
+	SigningScheme                SigningScheme
+	VerificationPlugin           string
+	VerificationPluginMinVersion string  // TODO: Implement SimVer structure
```
rgnote, line 50:
```suggestion
	VerificationPluginMinVersion string  // TODO: Implement SemVer structure
```

signer/jws.go
```diff
@@ -188,11 +196,19 @@ func populateProtectedHeaders(protectedHdr *jwsProtectedHeader, signInfo *Signer
...
 
 	signInfo.PayloadContentType = protectedHdr.ContentType
-	signInfo.SignedAttributes.SigningTime = protectedHdr.SigningTime.Truncate(time.Second)
+	signInfo.SignedAttributes.ExtendedAttributes = getExtendedAttributes(protectedHdr.ExtendedAttributes, protectedHdr.Critical)
+	signInfo.SigningScheme = protectedHdr.SigningScheme
+	signInfo.SignedAttributes.VerificationPlugin = protectedHdr.VerificationPlugin
+	signInfo.SignedAttributes.VerificationPluginMinVersion = protectedHdr.VerificationPluginMinVersion
 	if protectedHdr.Expiry != nil {
-		signInfo.SignedAttributes.Expiry = protectedHdr.Expiry.Truncate(time.Second)
+		signInfo.SignedAttributes.Expiry = *protectedHdr.Expiry
```
rgnote, line 204: nit: No need to use a pointer reference here and below

signer/jws.go
```diff
@@ -227,18 +259,19 @@ func validateCriticalHeaders(protectedHdr *jwsProtectedHeader) error {
 	// validate all required critical headers are present.
 	if len(mustMarkedCrit) != 0 {
 		// This is not taken care by VerifySignerInfo method
-		return MalformedSignatureError{"required headers not marked critical"}
+		keys := make([]string, 0, len(mustMarkedCrit))
+		for k := range mustMarkedCrit {
+			keys = append(keys, k)
+		}
+		return MalformedSignatureError{fmt.Sprintf("required headers not marked critical : %v", keys)}
```
rgnote, line 266:
```suggestion
		return MalformedSignatureError{fmt.Sprintf("these required headers are not marked as critical : %v", keys)}
```

signer/jws.go
```diff
@@ -301,12 +351,25 @@ type jwsProtectedHeader struct {
...
-	// The time at which the signature was generated.
-	SigningTime time.Time `json:"io.cncf.notary.signingTime"`
-
 	// The "best by use" time for the artifact, as defined by the signer.
 	Expiry *time.Time `json:"io.cncf.notary.expiry,omitempty"`
 
+	// Specifies the Notary v2 Signing Scheme used by the signature.
+	SigningScheme SigningScheme `json:"io.cncf.notary.signingScheme"`
+
+	// The time at which the signature was generated. only valid when signing scheme is SigningSchemeX509Default
```
rgnote, line 360:
```suggestion
	// The time at which the signature was generated. only valid when signing scheme is `notary.default.x509`
```

signer/jws.go
```diff
@@ -301,12 +351,25 @@ type jwsProtectedHeader struct {
...
+	SigningTime *time.Time `json:"io.cncf.notary.signingTime,omitempty"`
+
+	// The time at which the signature was generated. only valid when signing scheme is SigningSchemeX509SigningAuthority
```
rgnote, line 363:
```suggestion
	// The time at which the signature was generated. only valid when signing scheme is `notary.signingAuthority.x509`
```

signer/jws.go
```diff
@@ -301,12 +351,25 @@ type jwsProtectedHeader struct {
...
+	AuthenticSigningTime *time.Time `json:"io.cncf.notary.authenticSigningTime,omitempty"`
+
+	// VerificationPlugin specifies the name of the verification plugin that would be used to verify the signature.
```
rgnote, line 366:
```suggestion
	// VerificationPlugin specifies the name of the verification plugin that should be used to verify the signature.
```

signer/jws.go
```diff
@@ -301,12 +351,25 @@ type jwsProtectedHeader struct {
...
+	VerificationPlugin string `json:"io.cncf.notary.verificationPlugin,omitempty"`
+
+	// VerificationPluginMinVersion specifies the minimum version of the verification plugin that would be used to verify the signature.
```
rgnote, line 369:
```suggestion
	// VerificationPluginMinVersion specifies the minimum version of the verification plugin that should be used to verify the signature.
```

signer/jws.go
```diff
@@ -301,12 +351,25 @@ type jwsProtectedHeader struct {
...
+	// TODO: replace string with well defined struct for SimVer.
```
rgnote, line 370:
```suggestion
	// TODO: replace string with well defined struct for SemVer.
```

signer/signer.go
rgnote: We can accept a string here rather than a struct. The TODO part is validating that the SemVer string is valid

signer/signer.go
```diff
@@ -137,7 +150,10 @@ func validateSignerInfo(info *SignerInfo) error {
 	errorFunc := func(s string) error {
 		return MalformedSignatureError{msg: s}
 	}
-	if err := validate(info.Payload, info.PayloadContentType, info.SignedAttributes.SigningTime, info.SignedAttributes.Expiry, errorFunc); err != nil {
+
+	sAttr := info.SignedAttributes
+	if err := validate(info.Payload, info.PayloadContentType, sAttr.VerificationPlugin, sAttr.VerificationPluginMinVersion,
```
rgnote, line 155: Just pass the info struct rather than individual fields?

signer/jws.go
priteshbandi: Here we are copying data refereed by `protectedHdr.Expiry` pointer to `signInfo.SignedAttributes.Expiry`. 
`protectedHdr.Expiry` needs to be a pointer because expiry is optional field and shouldn't be present in signature if user haven't specified expiry.

signer/signer.go
priteshbandi: We are reusing validate function across signing and verification call. Both have different structure(SignRequest and SignerInfo)

signer/signer.go
priteshbandi: As we need to validate semver which has its own semantics and compare with other semver, IMO it would be better to have well defined struct and fail early.

priteshbandi: Since golang doesnt provides require library ([semver in x pkg](https://pkg.go.dev/golang.org/x/mod@v0.5.1/semver)) doest abide by https://semver.org/. Used regular expression to validate semver.

rgnote (approved):

signer/types.go
```diff
@@ -64,9 +64,17 @@ func (k KeySpec) SignatureAlgorithm() SignatureAlgorithm {
 	return ""
 }
 
+// SigningScheme formalizes the feature set (guarantees) provided by the signature.
+type SigningScheme string
+
+const (
+	SigningSchemeX509Default          SigningScheme = "notary.default.x509"
+	SigningSchemeX509SigningAuthority SigningScheme = "notary.signingAuthority.x509"
```
rgnote, line 72: These were changed to `notary.x509` and `notary.x509.signingAuthority` in the latest spec. See https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#standard-attributes

priteshbandi: Yep, waiting for @gokarnm to review and if there are any feedback will squash this renaming with other changes.

dtzar:
Per the [PR review criteria](https://github.com/notaryproject/notation/discussions/263) @notaryproject/notation-sub-group-maintainers - this is past a week old, BUT needs one more approval from @notaryproject/notation-code-reviewers, to be able to be merged.

shizhMSFT (approved):
LGTM

signer/jws.go
```diff
@@ -227,18 +259,19 @@ func validateCriticalHeaders(protectedHdr *jwsProtectedHeader) error {
 	// validate all required critical headers are present.
```
gokarnm, line 259: Nit, shouldn't this comment be in line 248? Here we throw an error if header is present but not marked critical.

signer/jws.go
```diff
@@ -209,11 +225,27 @@ func getExtendedAttributes(attrs map[string]interface{}, critical []string) []At
...
+	if len(protectedHdr.Critical) == 0 {
+		return MalformedSignatureError{"missing `crit` header"}
+	}
+
+	mustMarkedCrit := map[string]bool{headerKeySigningScheme: true}
 	if protectedHdr.Expiry != nil && !protectedHdr.Expiry.IsZero() {
 		mustMarkedCrit[headerKeyExpiry] = true
 	}
 
+	if protectedHdr.SigningScheme == SigningSchemeX509SigningAuthority {
```
gokarnm, line 237: For Signing scheme X509, presence of `AuthenticSigningTime` is invalid. And for X509 Signing Authority, presence of `SigningTime` is invalid. We should add those validations.

https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#signed-attributes

signer/signer.go
```diff
@@ -17,14 +19,17 @@ type SignerInfo struct {
 	SignatureAlgorithm SignatureAlgorithm
 	CertificateChain   []*x509.Certificate
 	Signature          []byte
+	SigningScheme      SigningScheme
```
gokarnm, line 22: Nit, I'd be preferable to organize these as signed and unsigned and have some comments.

```
// SignerInfo represents a parsed signature envelope that is agnostic to signature envelope format.
type SignerInfo struct {
     
       Payload            []byte

        // Signed attributes
        SignatureAlgorithm SignatureAlgorithm
	SigningScheme      SigningScheme
	PayloadContentType PayloadContentType
	SignedAttributes   SignedAttributes

       // The signature is generated over payload and signed attributes
       Signature          []byte

       // Unsigned attributes
       CertificateChain   []*x509.Certificate
	TimestampSignature []byte
	UnsignedAttributes UnsignedAttributes
	SignatureAlgorithm SignatureAlgorithm
}
```

signer/signer.go
```diff
@@ -189,12 +208,11 @@ func validateCertificateChain(certChain []*x509.Certificate, signTime time.Time,
 	return nil
 }
 
-func validate(payload []byte, payloadCty PayloadContentType, signTime, expTime time.Time, f func(string) error) error {
+func validate(payload []byte, payloadCty PayloadContentType, vPlugin, vPluginVersion string, signTime, expTime time.Time, scheme SigningScheme, f func(string) error) error {
```
gokarnm, line 211: Nit, I'd use full names for vPlugin, vPluginVersion

signer/jws.go
priteshbandi: Nop, this is the correct comment. We are doing two validations
1. Validate that all headers market as critical are present in the headers map.
2. Validate that all required headers(as per spec) are marked critical.

gokarnm (approved):
LGTM!

PR #20, opened 2022-07-28
fix hash function name according to the spec

chloeyin:
Signed-off-by: zaihaoyin <zaihaoyin@microsoft.com>

This PR fixes https://github.com/notaryproject/notation/issues/268.
According to the [plugin spec](https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md), the `hashAlgorithm` field of `GenerateSignatureRequest` should be one of the following strings.
```
  // Hash algorithm associated with the key spec, plugin must 
  // hash the payload using this hash algorithm
  "hashAlgorithm" : "SHA_256" | "SHA_384" | "SHA_512",
```
However, the current implementation uses `crypto.Hash.String()`, which will return a string like `SHA-256` and break the azure-kv plugin.
Also need to update notation-go and notation to make the plugin works.

### DIFF
--- a/signer/types.go
+++ b/signer/types.go
@@ -8,6 +8,9 @@ type SignatureMediaType string
 // SignatureAlgorithm lists supported signature algorithms.
 type SignatureAlgorithm string
 
+// HashAlgorithm algorithm associated with the key spec.
+type HashAlgorithm string
+
 // One of following supported specs
 // https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#algorithm-selection
 const (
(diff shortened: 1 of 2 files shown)


patrickzheng200:
@chloeyin We still need to change [notation-go](https://github.com/notaryproject/notation-go/blob/34be24db5095f86973546a5ae2ef631cbee6f3f9/signature/plugin.go#L257) right? "SHA-256" is actually not following the spec you mentioned above.

chloeyin:
> @chloeyin We still need to change [notation-go](https://github.com/notaryproject/notation-go/blob/34be24db5095f86973546a5ae2ef631cbee6f3f9/signature/plugin.go#L257) right? "SHA-256" is actually not following the spec you mentioned above.

Yes, we need to update notation-go and notation to make the plugin work.

patrickzheng200:
> > @chloeyin We still need to change [notation-go](https://github.com/notaryproject/notation-go/blob/34be24db5095f86973546a5ae2ef631cbee6f3f9/signature/plugin.go#L257) right? "SHA-256" is actually not following the spec you mentioned above.
> 
> Yes, we need to update notation-go and notation to make the plugin work.

And with this PR merged, the remaining fix is as simple as psp.keySpec.SignatureAlgorithm().Hash(), just removing the .String().

patrickzheng200:
LGTM

shizhMSFT (approved):
LGTM

gokarnm:
@chloeyin @shizhMSFT, I couldn't review this PR earlier. This isn't the correct fix.

The plugin spec is implemented by notation-go. notation-core-go does not know anything about plugin request/response types, which is where the `generate-signature.Request.hashAlgorithm` is defined. The fix involves translating hash alg names into ones required by plugin spec.

I'm actually in favor of the plugin spec being updated to use standard hash alg names, so we can avoid any translation at all.

https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md#generate-signature

PR #27, opened 2022-08-09
refactor: refine the envelope and signer to enhance extensiblity

binbin-li:
### What?
Refine the envelope and signer framework so that it can support jws and cose easily without coupling logics.


signature/errors.go
```diff
@@ -0,0 +1,51 @@
+package signature
+
+import "fmt"
+
+// MalformedSignatureError is used when Signature envelope is malformed.
+type MalformedSignatureError struct {
```
shizhMSFT, line 6: Let's refine those Errors in the next PR.

shizhMSFT (approved):
LGTM

signature/algorithm.go
```diff
@@ -11,12 +13,12 @@ type Algorithm int
...
-	AlgorithmPS512                      // RSASSA_PSS_SHA512
-	AlgorithmES256                      // ECDSA_SHA256
-	AlgorithmES384                      // ECDSA_SHA384
-	AlgorithmES512                      // ECDSA_SHA512
+	AlgorithmPS256 Algorithm = 1 + iota // RSASSA-PSS with SHA-256
+	AlgorithmPS384                      // RSASSA-PSS with SHA-384
+	AlgorithmPS512                      // RSASSA-PSS with SHA-512
+	AlgorithmES256                      // ECDSA on secp256r1 with SHA-256
+	AlgorithmES384                      // ECDSA on secp384r1 with SHA-384
+	AlgorithmES512                      // ECDSA on secp521r1 with SHA-512
```
chloeyin, line 21: Should we rename `AlgorithmES512` to `AlgorithmES521`?  Because the actual signing alg name is `secp521r1 `.

PR #28, opened 2022-08-09
feat: implemented COSE Envelope

patrickzheng200:
This PR targets the [notaryproject:cose](https://github.com/notaryproject/notation-core-go/tree/cose) branch.


signature/cose/cose.go
```diff
@@ -1,25 +1,155 @@
...
+	"crypto/x509"
+	"errors"
 	"fmt"
+	"io"
+	"time"
 
 	"github.com/notaryproject/notation-core-go/signature"
 	"github.com/notaryproject/notation-core-go/signature/internal/base"
+	nx509 "github.com/notaryproject/notation-core-go/x509"
+	cosepkg "github.com/veraison/go-cose"
```
shizhMSFT, line 17: Any reason to alias `go-cose` to `cosepkg` instead of `cose`?

Note: We can reference `cose` inside the package named `cose`.

signature/cose/cose.go
```diff
@@ -15,34 +26,472 @@ func init() {
...
+	if err != nil {
+		return nil, nil, err
+	}
+	certs, err := signer.CertificateChain()
+	if err != nil {
+		return nil, nil, err
+	}
+	var certChain [][]byte
+	for i, c := range certs {
+		certChain[i] = c.Raw
```
qmuntal, line 387: This line will always panic, `certChain` is not initialized, so indexed assignments panic.

qmuntal (changes requested):

signature/cose/cose.go
patrickzheng200: Code updated.

go.mod
```diff
@@ -3,3 +3,9 @@ module github.com/notaryproject/notation-core-go
 go 1.17
 
 require github.com/golang-jwt/jwt/v4 v4.4.1
+
+require (
+	github.com/fxamacker/cbor/v2 v2.4.0 // indirect
+	github.com/veraison/go-cose v1.0.0-rc.1 // indirect
```
shizhMSFT, line 9: This should be direct, shouldn't it?

signature/cose/cose.go
```diff
@@ -15,34 +26,524 @@ func init() {
...
+const (
+	headerKeyExpiry                       = "io.cncf.notary.expiry"
+	headerKeySigningTime                  = "io.cncf.notary.signingTime"
+	headerKeyAuthenticSigningTime         = "io.cncf.notary.authenticSigningTime"
+	headerKeySigningScheme                = "io.cncf.notary.signingScheme"
+	headerKeyVerificationPlugin           = "io.cncf.notary.verificationPlugin"
+	headerKeyVerificationPluginMinVersion = "io.cncf.notary.verificationPluginMinVersion"
+	headerKeyCrit                         = "crit"
+	headerKeyAlg                          = "alg"
+	headerKeyCty                          = "cty"
```
shizhMSFT, line 40: Those labels are integers instead of strings.

signature/cose/cose.go
```diff
@@ -15,34 +26,524 @@ func init() {
...
+	if e.coseEnvelope == nil {
+		return nil, &signature.MalformedSignatureError{Msg: "missing Cose signature envelope"}
+	}
+	if len(e.coseEnvelope.Payload) == 0 {
+		return nil, &signature.MalformedSignatureError{Msg: "missing payload"}
+	}
+	payload.Content = e.coseEnvelope.Payload
+	if cty, ok := e.coseEnvelope.Headers.Protected[cose.HeaderLabelContentType]; !ok {
+		return nil, &signature.MalformedSignatureError{Msg: "missing Content type"}
+	} else {
```
qmuntal, line 199: this `else` is not necessary as the main branch returns.

signature/cose/cose.go
```diff
@@ -15,34 +26,524 @@ func init() {
...
+	// parse signature
+	sig := e.coseEnvelope.Signature
+	if len(sig) == 0 {
+		return nil, &signature.MalformedSignatureError{Msg: "cose envelope missing signature"}
+	}
+	signInfo.Signature = sig
+
+	// parse unprotected headers
+	// x5chain
+	var certChain []*x509.Certificate
```
qmuntal, line 233: `certChain` definition can be moved some lines down, just before `range certs`.

qmuntal (changes requested):

signature/cose/cose.go
```diff
@@ -15,34 +26,524 @@ func init() {
...
+	return n
+}
+
 type envelope struct {
+	coseEnvelope *cose.Sign1Message
+}
+
+type pluginSigner struct {
+	signer signature.Signer
+}
```
shizhMSFT, line 76: We can just call it a signer instead of pluginSigner.

```suggestion
type signer struct {
	base signature.Signer
}
```

signature/cose/cose.go
```diff
@@ -15,34 +26,524 @@ func init() {
...
+	n := make(map[cose.Algorithm]signature.Algorithm, len(m))
+	for k, v := range m {
+		n[v] = k
+	}
+	return n
+}
+
 type envelope struct {
+	coseEnvelope *cose.Sign1Message
+}
```
shizhMSFT, line 72: We don't need to name it `cose` as it is already in the cose package.

```suggestion
type envelope cose.Sign1Message
```

or 

```suggestion
type envelope struct {
	base *cose.Sign1Message
}
```

signature/cose/cose.go
```diff
@@ -15,34 +26,524 @@ func init() {
...
+		default:
+			return 0, errors.New("key size not supported")
+		}
+	case signature.KeyTypeEC:
+		switch keySpec.Size {
+		case 256:
+			return signatureAlgCOSEAlgMap[signature.AlgorithmES256], nil
+		case 384:
+			return signatureAlgCOSEAlgMap[signature.AlgorithmES384], nil
+		case 512:
```
chloeyin, line 323: https://github.com/notaryproject/notation-core-go/blob/0553376c4e5d86681c177cbb5bfec37024dd69f6/signature/algorithm.go#L71-L80
Do we have a key size 512 if we use ecdsa?

signature/cose/cose.go
patrickzheng200: These are actually not used, will remove them. Cose is using go-cose's Headers instead.

signature/cose/cose.go
```diff
@@ -15,34 +26,524 @@ func init() {
...
+		}
+		protected[elm.Key] = elm.Value
+	}
+
+	return nil
+}
+
+func sign(req *signature.SignRequest, signer signature.Signer, msgToSign *cose.Sign1Message, alg cose.Algorithm) ([]byte, []*x509.Certificate, error) {
+	var coseSigner cose.Signer
+	var err error
```
chloeyin, line 369:
```suggestion
	var（
	    coseSigner cose.Signer
	    err error
	 ）
	    
```
I prefer defining them in a single bracket,

signature/cose/cose.go
patrickzheng200: This part of code is from Binbin, which is already merged.

signature/cose/cose.go
```diff
@@ -15,34 +26,524 @@ func init() {
...
+
+	// extended attributes
+	extendedAttributes := headers.Protected
+	delete(extendedAttributes, cose.HeaderLabelAlgorithm)
+	delete(extendedAttributes, cose.HeaderLabelContentType)
+	delete(extendedAttributes, cose.HeaderLabelCritical)
+	delete(extendedAttributes, headerKeySigningTime)
+	delete(extendedAttributes, headerKeyExpiry)
+	delete(extendedAttributes, headerKeyAuthenticSigningTime)
+	delete(extendedAttributes, headerKeySigningScheme)
```
binbin-li: probably we can move this block into `getExtendedAttributes`.

signature/cose/cose.go
```diff
@@ -15,34 +26,524 @@ func init() {
...
+	signInfo.SignedAttributes.ExtendedAttributes, err = getExtendedAttributes(protected, extendedAttributes)
+	if err != nil {
```
binbin-li, line 447: we can just return `err` early

signature/cose/cose.go
```diff
@@ -15,34 +26,524 @@ func init() {
...
+		return nil, err
+	}
 	return &base.Envelope{
-		Envelope: &envelope{},
+		Envelope: &envelope{coseEnvelope: &coseEnvelope},
 		Raw:      envelopeBytes,
 	}, nil
 }
+
+func validateCertificateChain(certChain []*x509.Certificate, signTime time.Time, expectedAlg cose.Algorithm, f func(string) error) error {
```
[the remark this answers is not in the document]
patrickzheng200: > we have duplicate cert validation method, just leave a comment so we can refactor it in another PR.

Just to check, your version is in another package and is not exported right? So looks like I'm not able to directly use it in cose package at the moment.

signature/cose/cose.go
```diff
@@ -15,34 +26,524 @@ func init() {
...
+		protected[headerKeyExpiry] = uint(req.Expiry.Unix())
+	}
+	switch req.SigningScheme {
+	case signature.SigningSchemeX509:
+		protected[headerKeySigningTime] = uint(req.SigningTime.Unix())
+	case signature.SigningSchemeX509SigningAuthority:
+		crit = append(crit, headerKeyAuthenticSigningTime)
+		protected[headerKeyAuthenticSigningTime] = uint(req.SigningTime.Unix())
+	default:
+		return errors.New("SigningScheme: require notary.x509 or notary.x509.signingAuthority")
```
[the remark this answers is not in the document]
patrickzheng200: Shiwei said to separate SigningScheme for a future PR, so I removed SigningScheme related logic for now.

signature/cose/cose.go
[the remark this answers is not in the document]
binbin-li: right, we can address it once the workflow works.

binbin-li (approved):
lgtm

PR #34, opened 2022-08-12
feat: sign by JWT package & add signingScheme, crit

JeyJeyGao:
1. JWT sign replace custom sign
2. add SigningScheme, crit


signature/jws/types.go
```diff
@@ -52,3 +91,22 @@ type JwsEnvelope struct {
...
+var signatureAlgJWSAlgMap = map[signature.Algorithm]string{
+	signature.AlgorithmPS256: "PS256",
+	signature.AlgorithmPS384: "PS384",
+	signature.AlgorithmPS512: "PS512",
+	signature.AlgorithmES256: "ES256",
+	signature.AlgorithmES384: "ES384",
+	signature.AlgorithmES512: "ES512",
+}
+
+var jwsAlgSignatureAlgMap = reverseMap(signatureAlgJWSAlgMap)
```
chloeyin, line 104: Do we need a `reverseMap`? Where do you use the `jwsAlgSignatureAlgMap`

signature/jws/envelope.go
```diff
@@ -0,0 +1,190 @@
...
+	if err != nil {
+		return nil, nil, &signature.MalformedSignatureError{Msg: "malformed leaf certificate"}
+	}
+
+	// verify JWT
+	compact := strings.Join([]string{
+		e.internalEnvelope.Protected,
+		e.internalEnvelope.Payload,
+		e.internalEnvelope.Signature}, ".")
+	err = verifyJWT(compact, cert)
```
binbin-li, line 95: nit: line 95 can be moved in the if statement.

signature/jws/envelope.go
```diff
@@ -0,0 +1,190 @@
...
+		return nil, nil, err
+	}
+
+	// extract payload and signer info
+	payload, err := e.Payload()
+	if err != nil {
+		return nil, nil, err
+	}
+	signerInfo, err := e.SignerInfo()
+	return payload, signerInfo, err
```
binbin-li, line 106: do we want to return a valid payload though the err is not nil?

signature/jws/types.go
```diff
@@ -1,6 +1,36 @@
...
+	headerKeyCrit                         = "crit"
+	headerKeyAlg                          = "alg"
+	headerKeyCty                          = "cty"
+	headerKeyAuthenticSigningTime         = "io.cncf.notary.authenticSigningTime"
+	headerKeySigningScheme                = "io.cncf.notary.signingScheme"
+	headerKeyVerificationPlugin           = "io.cncf.notary.verificationPlugin"
+	headerKeyVerificationPluginMinVersion = "io.cncf.notary.verificationPluginMinVersion"
+)
+
+func getHeaderKeys() []string {
```
chloeyin, line 21:
```suggestion
	headerKeys := []string{}
```
I prefer to use a variable instead because I don't see why we need a function. Are we going to add some logic to this function in the future?

signature/jws/jwt.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return "", err
+	}
+	return getJWSAlgo(keySpec.SignatureAlgorithm())
+}
+
+func (s *remoteSigningMethod) Alg() string {
+	alg, err := getJwsAlgorithm(s.signer)
+	if err != nil {
+		panic(err)
```
chloeyin, line 85: Can we return an error instead of calling panic and crash the whole process?

signature/jws/jwt.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/notaryproject/notation-core-go/signature"
+)
+
+// remoteSigningMethod wraps the external signer to be a jwt.SigningMethod
+type remoteSigningMethod struct {
+	jwt.SigningMethod
+	signer signature.Signer
+}
+
+func (s *remoteSigningMethod) Sign(signingString string, key interface{}) (string, error) {
```
binbin-li, line 17: nit: missing comment

signature/jws/types.go
JeyJeyGao: Yes. When we call SignerInfo(), we need to parse protected header to get the string format algorithm and convert it to be signature.Algorithm and save in SignInfo struct.

signature/jws/jwt.go
```diff
@@ -0,0 +1,109 @@
...
+	"crypto/x509"
+	"fmt"
+
+	"github.com/golang-jwt/jwt/v4"
+	"github.com/notaryproject/notation-core-go/signature"
+)
+
+// remoteSigningMethod wraps the external signer to be a jwt.SigningMethod
+type remoteSigningMethod struct {
+	jwt.SigningMethod
```
binbin-li, line 13: Do we need to implement `Alg()` method?

signature/jws/types.go
JeyJeyGao: No logic here. Just for save memory. 
Changed it to be a variable.

signature/jws/jwt.go
JeyJeyGao: I cannot return an error, because remoteSigningMethod implements this interface in jwt package:

> // SigningMethod can be used add new methods for signing or verifying tokens.
type SigningMethod interface {
	Verify(signingString, signature string, key interface{}) error // Returns nil if signature is valid
	Sign(signingString string, key interface{}) (string, error)    // Returns encoded signature or error
	Alg() string                                                   // returns the alg identifier for this method (example: 'HS256')
}

signature/jws/envelope.go
JeyJeyGao: Sure.

signature/jws/envelope.go
JeyJeyGao: Updated.

signature/jws/jwt.go
JeyJeyGao: Good question. No, we don't need. Removed the `Alg()` method.

signature/jws/jwt.go
JeyJeyGao: Added.

main.go
```diff
@@ -0,0 +1,14 @@
...
+func main() {
+	a := [][]byte{}
+	var c interface{}
+	c = a
+	k, ok := c.([][]byte)
+	if ok {
+		fmt.Println("OK")
+	}
+	fmt.Println(k)
+}
```
patrickzheng200, line 14: Looks like this is for testing?

signature/jws/envelope_test.go
```diff
@@ -57,6 +45,7 @@ func getSignReq() (*signature.SignRequest, error) {
 		Expiry:                   time.Now().Add(time.Hour),
 		ExtendedSignedAttributes: nil,
 		SigningAgent:             "Notation/1.0.0",
+		SigningScheme:            "notary.x509",
```
patrickzheng200, line 48: we can pass signing scheme as an argument of getSignReq(), so that we can use a loop to test both SigningSchemes.

signature/types.go
```diff
@@ -29,12 +39,15 @@ type Attribute struct {
...
-	SigningAgent             string
+	Payload                      Payload
+	Signer                       Signer
+	SigningTime                  time.Time
+	Expiry                       time.Time
+	ExtendedSignedAttributes     []Attribute
+	SigningAgent                 string
+	SigningScheme                SigningScheme
+	VerificationPlugin           string
+	VerificationPluginMinVersion string
```
patrickzheng200, line 50: Same here.

signature/types.go
```diff
@@ -5,14 +5,24 @@ import (
...
 
 // SignedAttributes represents signed metadata in the Signature envelope
 type SignedAttributes struct {
-	SigningTime        time.Time
-	Expiry             time.Time
-	ExtendedAttributes []Attribute
+	SigningTime                  time.Time
+	Expiry                       time.Time
+	VerificationPlugin           string
+	VerificationPluginMinVersion string
```
patrickzheng200, line 24: Do we need these two in notation-core-go?

main.go
JeyJeyGao: Removed.

signature/types.go
JeyJeyGao: Removed.

signature/jws/envelope_test.go
JeyJeyGao: Good idea. updated.

patrickzheng200 (approved):
LGTM

PR #40, opened 2022-08-16
feat: validate key and cert while creating local signer

binbin-li:
### What?

Validate if the privateKey and leaf certificate match while creating local signers.


signature/signer.go
```diff
@@ -49,6 +57,38 @@ func NewLocalSigner(certs []*x509.Certificate, key crypto.PrivateKey) (LocalSign
...
+
+	switch publicKey.(type) {
+	case *rsa.PublicKey:
+		switch privateKey.(type) {
+		case *rsa.PublicKey:
+			key, ok := privateKeySigner.Public().(*rsa.PrivateKey)
+			if !ok {
+				return false
+			}
+			return key.Equal(publicKey.(*rsa.PublicKey))
```
patrickzheng200, line 72: An alternative way of checking the public key matches the private key is: [tls](https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/crypto/tls/tls.go;l=310)

binbin-li: good call, updated.

patrickzheng200 (approved):
LGTM

PR #43, opened 2022-08-17
test: add signature unit tests

binbin-li:
### What?

1. Add unit tests for `signature` submodule.
2. Add unit tests for internal base envelope submodule.

Test coverage: 
image

image


signature/algorithm_test.go
```diff
@@ -0,0 +1,199 @@
...
+		cert      *x509.Certificate
+		expect    KeySpec
+		expectErr bool
+	}{
+		{
+			name: "RSA 3072",
+			cert: testhelper.GetRSALeafCertificate().Cert,
+			expect: KeySpec{
+				Type: KeyTypeRSA,
+				Size: 3072,
```
chloeyin, line 78: For valid cases, can we enumerate all possible keySpec? eg: RSA 2048, RSA 4096
```suggestion
               // Same for ecdsa,  just an example
		for _, k := range []int{2048, 3072, 4096} {
			rsaRoot := testhelper.GetRSARootCertificate()
		        pk, _ := rsa.GenerateKey(rand.Reader, k)
		        certTuple := testhelper.GetRSACertTupleWithPK(pk, "TestWithRSA_"+strconv.Itoa(pk.Size()), &rsaRoot)
		        tests = append(tests, []struct{
		                name: "RSA"+strconv.Itoa(k),
		                cert: certTuple.Cert,
		                expect: KeySpec{
		                      Type: KeyTypeRSA,
		                      SIze: k,
		                },
		                expectErr: false,
		        }
		 }
```

signature/algorithm_test.go
```diff
@@ -0,0 +1,199 @@
...
+			expect: AlgorithmPS512,
+		},
+		{
+			name: "Unsupported key spec",
+			keySpec: KeySpec{
+				Type: 0,
+				Size: 0,
+			},
+			expect: 0,
+		},
```
chloeyin, line 188: Open a discussion: Can we find a way to use a for loop to enumerate all kinds of keySize and keyType to avoid hard code all cases?

JeyJeyGao: write a helper to generate test data?

binbin-li: I feel even we create a slice or a map for keySize and keyType, we still need to hard code the expected result and the test name. It would be more readable in current way instead of splitting keySpec and test name/expected result into different slices or maps.

signature/algorithm_test.go
binbin-li: Good call, I'll add all valid cases.

chloeyin (approved):

patrickzheng200 (approved):
LGTM

PR #44, opened 2022-08-18
fix: update envelope.Payload() logic

JeyJeyGao:
1. Payload() returns the raw payload context instead of base64 encoded data
2. move extract algorithm step from Sign() to getSignedAttrs()
3. wrap the parse error in a MalformedSignatureError

### DIFF
--- a/signature/jws/envelope.go
+++ b/signature/jws/envelope.go
@@ -36,7 +36,7 @@ func ParseEnvelope(envelopeBytes []byte) (signature.Envelope, error) {
 	var e jwsEnvelope
 	err := json.Unmarshal(envelopeBytes, &e)
 	if err != nil {
-		return nil, err
+		return nil, &signature.MalformedSignatureError{Msg: err.Error()}
 	}
 	return &base.Envelope{
 		Envelope: &envelope{internalEnvelope: &e},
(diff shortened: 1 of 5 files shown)


signature/jws/envelope.go
```diff
@@ -129,12 +124,29 @@ func (e *envelope) Payload() (*signature.Payload, error) {
...
 	if err != nil {
 		return nil, err
 	}
+
+	// convert JWS to JWT
+	tokenString := compactJWS(e.internalEnvelope)
+
+	// parse JWT to get payload context
+	parser := jwt.NewParser(
+		jwt.WithValidMethods([]string{"PS256", "PS384", "PS512", "ES256", "ES384", "ES512"}),
```
chloeyin, line 138: Can we use a `validMethod` slice instead of hard code string in the code

chloeyin, line 138:
```suggestion
               var validMethods = []string{"PS256", "PS384", "PS512", "ES256", "ES384", "ES512"}
		jwt.WithValidMethods(validMethods),
```
Can we use a `validMethod` slice instead of hard code string in the code

JeyJeyGao: Good idea. Updated.

signature/jws/envelope_test.go
```diff
@@ -89,7 +90,7 @@ func Test_envelope_Verify_failed(t *testing.T) {
 	if err != nil {
 		t.Fatal(t)
 	}
-
+	fmt.Printf("encoded: %v\n", string(encoded))
```
chloeyin, line 93:
```suggestion
	t.Logf("encoded: %v\n", string(encoded))
```

JeyJeyGao: Thanks! Removed it.

chloeyin (approved):

PR #45, opened 2022-08-18
Updated COSE envelope unit tests

patrickzheng200:
Test Coverage:
image

Signed-off-by: Patrick Zheng [patrickzheng@microsoft.com](mailto:patrickzheng@microsoft.com)


signature/cose/envelope_test.go
```diff
@@ -79,144 +39,177 @@ func TestParseEnvelopeError(t *testing.T) {
...
+				size = []int{256, 384, 521}
 			}
-		})
+			for _, size := range size {
+				t.Run(fmt.Sprintf("with %s scheme, %v keyType, %v keySize when all arguments are present", signingScheme, keyType, size), func(t *testing.T) {
+					signRequest, err := newSignRequest(signingScheme, keyType, size)
+					if err != nil {
+						t.Fatalf("newSignRequest() failed. Error = %s", err)
+					}
+					_, err = env.Sign(signRequest)
```
binbin-li, line 58: should we do some checking on the returned signature envelope bytes?

signature/cose/envelope_test.go
```diff
@@ -79,144 +39,177 @@ func TestParseEnvelopeError(t *testing.T) {
...
 	}
 }
 
-// Tests various error scenarios around generating a COSE envelope
 func TestSignErrors(t *testing.T) {
-	env := NewEnvelope()
-	t.Run("when Payload is absent", func(t *testing.T) {
+	env := envelope{}
+	// Testing getSigner()
+	t.Run("errorLocalSigner: when getSigner has privateKeyError", func(t *testing.T) {
```
binbin-li, line 127: looks like we have a lot of similar tests, maybe we can try table-driven tests like this: https://yourbasic.org/golang/table-driven-unit-test/

signature/cose/envelope_test.go
```diff
@@ -227,73 +220,679 @@ func TestSignErrors(t *testing.T) {
...
+					if err != nil {
+						t.Fatalf("newSignRequest() failed. Error = %s", err)
+					}
+					encoded, err := env.Sign(signRequest)
+					if err != nil || len(encoded) == 0 {
+						t.Fatalf("Sign() faild. Error = %s", err)
+					}
+
+					// Verify using the same envelope struct
+					_, _, err = env.Verify()
```
binbin-li, line 644: should we create a new `env` by parsing the `encoded`?

signature/cose/envelope_test.go
```diff
@@ -227,73 +220,679 @@ func TestSignErrors(t *testing.T) {
...
+			return signature.NewLocalSigner(certs, leafCertTuple.PrivateKey)
+		default:
+			return nil, fmt.Errorf("key size not supported")
+		}
+	default:
+		return nil, fmt.Errorf("keyType not supported")
+	}
+}
+
+func getRSACertTuple(size int) testhelper.RSACertTuple {
```
binbin-li, line 744: discussion: which is a better location for the below methods, `testhelper` or this file.

signature/cose/envelope_test.go
patrickzheng200: This one is intended to use the same envelope as this is the situation when we sign in notation-go. TestSignAndParseVerify does create a new env by UnmarshalCBOR.

signature/cose/envelope_test.go
patrickzheng200: code updated

signature/cose/envelope_test.go
patrickzheng200: nice catch. Moved these two funcs into testhelper.

signature/cose/envelope_test.go
patrickzheng200: Looks like this pattern does not do much help in my case because I need to manipulate different fields of signRequest. Changing signer and SigningScheme do not share the same t.Run(), so it's necessary to create more than one test structs.

patrickzheng200: > looks like we have a lot of similar tests, maybe we can try table-driven tests like this: https://yourbasic.org/golang/table-driven-unit-test/

In my case, I need to manipulate different fields of signRequest. For example, one test changes the signer; another one changes the SigningScheme. To reuse the same t.Run(), it's necessary to include a func field in the test struct, this in some extent decreases the readability of the code (there will be a very long test struct slice with a func embedded in each of the test item).

binbin-li (approved):
lgtm

PR #50, opened 2022-08-23
updated COSE envelope for the certification chain changes

patrickzheng200:
Signed-off-by: Patrick Zheng [patrickzheng@microsoft.com](mailto:patrickzheng@microsoft.com)

### DIFF
--- a/signature/cose/envelope.go
+++ b/signature/cose/envelope.go
@@ -4,7 +4,6 @@ import (
 	"crypto"
 	"crypto/rand"
 	"crypto/x509"
-	"errors"
 	"fmt"
 	"io"
 	"strconv"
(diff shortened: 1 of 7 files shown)


signature/cose/envelope.go
```diff
@@ -382,22 +421,23 @@ func generateProtectedHeaders(req *signature.SignRequest, protected cose.Protect
...
 	// signing agent
 	unprotected[headerLabelSigningAgent] = req.SigningAgent
 
 	// certChain
-	certs, err := req.Signer.CertificateChain()
-	if err != nil {
-		return err
+	var certs []*x509.Certificate
+	switch s := signer.(type) {
+	case *remoteSigner:
```
binbin-li, line 431: nit: `case *remoteSigner, *localSigner:`

binbin-li (approved):
one minor comment, lgtm overall

signature/cose/envelope.go
patrickzheng200: @binbin-li I actually have tried that but got the following error: s.certs undefined (type "github.com/veraison/go-cose".Signer has no field or method certs).
variable s becomes cose.Signer type, and it does not have certs field defined, so looks like I have to separate them into 2 cases.

binbin-li: gotcha, good to know that.

PR #51, opened 2022-08-23
fix: fix the certs validation in envelope.Sign()

binbin-li:
### What?

Fix the workflow in envelope.Sign() so that we validate the certificate chain after signing correctly.


patrickzheng200 (approved):
LGTM

PR #62, opened 2022-08-31
fix: update jwt.Claims

JeyJeyGao:
jwt.Claims is used as jws's payload which is required to be JSON compatible object.


signature/jws/envelope.go
```diff
@@ -57,8 +58,16 @@ func (e *envelope) Sign(req *signature.SignRequest) ([]byte, error) {
 		return nil, err
 	}
 
+	// parse payload as jwt.MapClaims
```
patrickzheng200, line 61: Let's add a Reference URL to jwt.MapClaims in the comment.

signature/jws/envelope.go
```diff
@@ -57,8 +58,16 @@ func (e *envelope) Sign(req *signature.SignRequest) ([]byte, error) {
...
+	var payload jwt.MapClaims
+	err = json.Unmarshal(req.Payload.Content, &payload)
```
binbin-li, line 63: nit: we can combine line 63 and 64

signature/jws/envelope_test.go
```diff
@@ -563,7 +563,7 @@ func TestPayload(t *testing.T) {
 		checkNoError(t, err)
 
 		_, err = newEnv.Payload()
-		checkErrorEqual(t, "illegal base64 data at input byte 476", err.Error())
+		checkErrorEqual(t, "payload error: illegal base64 data at input byte 288", err.Error())
```
chloeyin, line 566: I found this 288 strange... Can we do a fuzz check. For example, checking the error message contains the keyword

signature/jws/envelope.go
JeyJeyGao: Added.

signature/jws/envelope.go
JeyJeyGao: Good idea.

signature/jws/envelope_test.go
JeyJeyGao: Updated.
```
func checkErrorEqual(t *testing.T, want, got string) {
	if !strings.Contains(got, want) {
		t.Fatalf("want: %v got: %v\n", want, got)
	}
}
```

patrickzheng200 (approved):
lgtm

PR #59, opened 2022-08-30
refactor: refactor SignatureProvider to Signer

binbin-li:
### What?

Background can be checked out in https://github.com/notaryproject/notation/discussions/278

1. Created a new `signature` submodule to refactor the original `signer` submodule. 
2. Renamed the original `SignatureProvider` to `Signer`. 
3. Created `LocalSigner` to support native signing implementation provided by underlying crypto libraries, e.g. go-cose, go-jwt.
4. Other refactoring on the algorithm and keySpec definitions.

### Test?
Added corresponding unit tests.

Signed-off-by: Binbin Li <libinbin@microsoft.com>

Issue #58, opened 2022-08-29
Refactor on SignRequest


binbin-li:
# Background
In the [PR](https://github.com/notaryproject/notation-core-go/pull/17/files#diff-29c2352f30841cae5d796cb41c2cb6b69dd1b5a2ff365c1fea32aaf2efae0b58R45), we introduced `VerificationPlugin` and `VerificationPluginMinVersion` to the `SignRequest` to support [validate ](https://github.com/notaryproject/notation-core-go/pull/17/files#diff-29c2352f30841cae5d796cb41c2cb6b69dd1b5a2ff365c1fea32aaf2efae0b58R236) them in the notation-core-go.

# Problem
1. As specified in specs: [Extended attributes for Notation Plugins](https://github.com/notaryproject/notaryproject/blob/cose-envelope/signature-specification.md#extended-attributes-for-notation-plugins) and [Extended Protected Headers for Notation Plugins](https://github.com/notaryproject/notaryproject/blob/main/signature-envelope-jws.md#extended-protected-headers-for-notation-plugins), `VerificationPlugin` and `VerificationPluginMinVersion` should be placed inside the `ExtendedSignedAttrs` of `SignRequest`.
2. For the validation, according to [Verification workflow using plugin](https://github.com/notaryproject/notaryproject/blob/cose-envelope/specs/plugin-extensibility.md#verification-workflow-using-plugin), we can see that it's reasonable to group version validation and capability validation together once we retrieve the plugin metadata. However, current implementation puts[ version validation](https://github.com/notaryproject/notation-core-go/blob/main/signer/signer.go#L241) in notation-core-go and [capability validation](https://github.com/notaryproject/notation-go/blob/main/verification/verifier.go#L163) in notation-go. Besides it, part of [version validation](https://github.com/notaryproject/notation-go/blob/main/verification/verifier.go#L160) is placed in notation-go which made reviewers confused on the validation workflow.

# Proposed refactoring
1. Move `VerificationPlugin` and `VerificationPluginMinVersion` to `ExtendedSignedAttrs` of `SignRequest`.
2. Move all version validation logics to notation-go, specifically [version validation](https://github.com/notaryproject/notation-go/blob/main/verification/verifier.go#L160)

binbin-li:
The first refactoring is not valid now. And the second one is done. Closing the issue.

PR #26, opened 2022-08-08
COSE envelope implementation updates

shizhMSFT:
Tracking the merger of the [cose](https://github.com/notaryproject/notation-core-go/tree/cose) branch into `main`.

PR #68, opened 2022-09-06
refactor: refactor envelope and signer to support cose

binbin-li:
### What?

Background can be checked out in https://github.com/notaryproject/notation/discussions/278
Same as PR #59, reopened for additional feedback.

1. Created a new `signature` submodule to refactor the original `signer` submodule. 
2. Created `Envelope` interface.
3. Renamed the original `SignatureProvider` to `Signer`. 
4. Created `LocalSigner` to support native signing implementation provided by underlying crypto libraries, e.g. go-cose, go-jwt.
5. Created `internal.base.Envelope` struct to replace the orginal `SignatureEnvelope` struct.
6. Other refactoring on the algorithm and keySpec definitions.

### Test?
Added corresponding unit tests.

### Notes
cose and jws folders are left empty for now which will be addressed in following PRs.

### DIFF
--- a/signature/envelope.go
+++ b/signature/envelope.go
@@ -1,0 +1,83 @@
+package signature
+
+import "fmt"
+
+// Envelope provides functions to basic functions to manipulate signatures.
+type Envelope interface {
+	// Sign generates and sign the envelope according to the sign request.
+	Sign(req *SignRequest) ([]byte, error)
+
+	// Verify verifies the envelope and returns its enclosed payload and signer
+	// info.
+	Verify() (*Payload, *SignerInfo, error)
+
+	// Payload returns the payload of the envelope.
+	// Payload is trusted only after the successful call to `Verify()`.
+	Payload() (*Payload, error)
+
+	// SignerInfo returns the signer information of the envelope.
+	// SignerInfo is trusted only after the successful call to `Verify()`.
+	SignerInfo() (*SignerInfo, error)
+}
+
+// NewEnvelopeFunc defines a function to create a new Envelope.
+type NewEnvelopeFunc func() Envelope
+
+// ParseEnvelopeFunc defines a function that takes envelope bytes to create
+// an Envelope.
+type ParseEnvelopeFunc func([]byte) (Envelope, error)
+
+// envelopeFunc wraps functions to create and parsenew envelopes.
+type envelopeFunc struct {
+	newFunc   NewEnvelopeFunc
+	parseFunc ParseEnvelopeFunc
+}
+
+// envelopeFuncs maps envelope media type to corresponding constructors and 
+// parsers.
+var envelopeFuncs = make(map[string]envelopeFunc)
+
+// RegisterEnvelopeType registers newFunc and parseFunc for the given mediaType.
+// Those functions are intended to be called when creating a new envelope.
+// It will be called while inializing the built-in envelopes(JWS/COSE).
+func RegisterEnvelopeType(mediaType string, newFunc NewEnvelopeFunc, parseFunc ParseEnvelopeFunc) error {
+	if newFunc == nil || parseFunc == nil {
+		return fmt.Errorf("required functions not provided")
+	}
+
+	envelopeFuncs[mediaType] = envelopeFunc{
+		newFunc:   newFunc,
+		parseFunc: parseFunc,
+	}
+	return nil
+}
+
+// RegisteredEnvelopeTypes lists registered envelope media types.
+func RegisteredEnvelopeTypes() []string {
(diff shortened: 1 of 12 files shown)


shizhMSFT (approved):
LGTM. The full picture can be found in the parent PR #26.

gokarnm (changes requested):
Added some feedback, will go through usages in #26 tomorrow to add any additional feedback.

signature/types.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// Critical marks the attribute that MUST be processed by a verifier.
+	Critical bool
+
+	// Value is the value of the attribute.
+	Value interface{}
+}
+
+// SignRequest is used to generate Signature.
+type SignRequest struct {
```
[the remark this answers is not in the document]
binbin-li: I feel we can move it to extended attributes as spec mentioned. https://github.com/notaryproject/notation-core-go/issues/58

signature/envelope.go
```diff
@@ -0,0 +1,96 @@
...
+	// Sign generates and sign the envelope according to the sign request.
+	Sign(req *SignRequest) ([]byte, error)
+
+	// Verify verifies the envelope and returns its enclosed payload and signer
+	// info.
+	Verify() (*Payload, *SignerInfo, error)
+
+	// Payload returns the payload of the envelope.
+	// Payload is trusted only after the successful call to `Verify()`.
+	Payload() (*Payload, error)
```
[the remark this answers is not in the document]
shizhMSFT: It gives the ability to extract the payload without verification. It is useful when the signature is verified elsewhere.

signature/types.go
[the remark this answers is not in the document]
shizhMSFT: `notation-core-go` should have plugin stuffs decoupled. That is, all things related to plugin should not be in the `notation-core-go` repo.

signature/types.go
```diff
@@ -0,0 +1,133 @@
...
+	// notary.x509 signing scheme.
+	SigningSchemeX509 SigningScheme = "notary.x509"
+
+	// notary.x509.signingAuthority schema.
+	SigningSchemeX509SigningAuthority SigningScheme = "notary.x509.signingAuthority"
+)
+
+// SignedAttributes represents signed metadata in the signature envelope.
+// Reference: https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#signed-attributes
+type SignedAttributes struct {
```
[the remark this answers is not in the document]
shizhMSFT: `authentic signing time` should be in another PR if we need it explicitly since it is not there in the `main` branch: https://github.com/notaryproject/notation-core-go/blob/main/signer/signer.go#L45

signature/envelope.go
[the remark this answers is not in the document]
gokarnm: That reason makes sense. Is there an example of how this would apply in notation?

signature/types.go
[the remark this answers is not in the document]
gokarnm: There are different levels of decoupling here. notation-go implements the plugin interactions, but notation-core-go signature envelope can be plugin aware by having typed fields for plugin attributes which are optional. This allows consumers of notation-core-go to use verification plugin attributes easily without each consumer having to know the attribute key names.

signature/envelope.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// SignerInfo returns the signer information of the envelope.
+	// SignerInfo is trusted only after the successful call to `Verify()`.
+	SignerInfo() (*SignerInfo, error)
```
gokarnm, line 30: SignerInfo may not be best name, consider SignatureInfo or similar. Payload should be part of this struct.

vaninrao10: I was thinking in the same line for Payload() (*Payload, error) - Payload is any payload in general, so should be call it Payload Envelope().

shizhMSFT: How about
```go
Content() (*EnvelopeContent, error)
```

shizhMSFT: or
```
Data() (*EnvelopeData, error)
```

shizhMSFT: where `EnvelopeConent` is a mimic of `SignedData` in CMS.

For CMS structs, see https://github.com/notaryproject/notation-core-go/blob/main/internal/crypto/cms/cms.go

binbin-li: thanks for the advice. I feel `Content()` sounds better to me.

vaninrao10: I vote for content. Like it !! Thank You

PR #73, opened 2022-09-09
refactor: refactor envelope and signer to support cose

binbin-li:
### What?

Background can be checked out in https://github.com/notaryproject/notation/discussions/278

1. Created `Envelope` interface.
2. Renamed the original `SignatureProvider` to `Signer`. 
3. Created `LocalSigner` to support native signing implementation provided by underlying crypto libraries, e.g. go-cose, go-jwt.
4. Created `internal.base.Envelope` struct to replace the orginal `SignatureEnvelope` struct.
5. Renamed `SignerInfo` to `EnvelopeContent` to better reflect its functionality.
6. Other refactoring on the algorithm and keySpec definitions.

### Test?
Added corresponding unit tests. 

### Notes
The compilation failed right now, which should be addressed in the next PR.

### DIFF
--- a/signature/envelope.go
+++ b/signature/envelope.go
@@ -44,7 +44,7 @@ type envelopeFunc struct {
 
 // envelopeFuncs maps envelope media type to corresponding constructors and
 // parsers.
-var envelopeFuncs sync.Map
+var envelopeFuncs sync.Map // map[string]envelopeFunc
 
 // RegisterEnvelopeType registers newFunc and parseFunc for the given mediaType.
 // Those functions are intended to be called when creating a new envelope.
--- a/signature/envelope.go
+++ b/signature/envelope.go
@@ -1,0 +1,92 @@
+// Package signature provides operations for types that implement
+// signature.Envelope or signature.Signer.
+//
+// An Envelope is a structure that creates and verifies a signature using the
+// specified signing algorithm with required validation. To register a new
+// envelope, call RegisterEnvelopeType first during the initialization.
+//
+// A Signer is a structure used to sign payload generated after signature
+// envelope created. The underlying signing logic is provided by the underlying
+// local crypto library or the external signing plugin.
+package signature
+
+import "fmt"
+
+// Envelope provides functions to basic functions to manipulate signatures.
+type Envelope interface {
+	// Sign generates and sign the envelope according to the sign request.
+	Sign(req *SignRequest) ([]byte, error)
+
+	// Verify verifies the envelope and returns its enclosed payload and signer
+	// info.
+	Verify() (*EnvelopeContent, error)
+
+	// Content returns the payload and signer information of the envelope.
+	// Content is trusted only after the successful call to `Verify()`.
+	Content() (*EnvelopeContent, error)
+}
+
+// NewEnvelopeFunc defines a function to create a new Envelope.
+type NewEnvelopeFunc func() Envelope
+
+// ParseEnvelopeFunc defines a function that takes envelope bytes to create
+// an Envelope.
+type ParseEnvelopeFunc func([]byte) (Envelope, error)
+
+// envelopeFunc wraps functions to create and parsenew envelopes.
+type envelopeFunc struct {
+	newFunc   NewEnvelopeFunc
+	parseFunc ParseEnvelopeFunc
+}
+
+// envelopeFuncs maps envelope media type to corresponding constructors and
+// parsers.
+var envelopeFuncs map[string]envelopeFunc
+
+// RegisterEnvelopeType registers newFunc and parseFunc for the given mediaType.
+// Those functions are intended to be called when creating a new envelope.
+// It will be called while inializing the built-in envelopes(JWS/COSE).
+func RegisterEnvelopeType(mediaType string, newFunc NewEnvelopeFunc, parseFunc ParseEnvelopeFunc) error {
+	if newFunc == nil || parseFunc == nil {
+		return fmt.Errorf("required functions not provided")
+	}
+	if envelopeFuncs == nil {
+		envelopeFuncs = make(map[string]envelopeFunc)
+	}
+
+	envelopeFuncs[mediaType] = envelopeFunc{
+		newFunc:   newFunc,
+		parseFunc: parseFunc,
+	}
+	return nil
+}
+
+// RegisteredEnvelopeTypes lists registered envelope media types.
+func RegisteredEnvelopeTypes() []string {
+	var types []string
+
+	for envelopeType := range envelopeFuncs {
+		types = append(types, envelopeType)
+	}
+
+	return types
+}
+
+// NewEnvelope generates an envelope of given media type.
+func NewEnvelope(mediaType string) (Envelope, error) {
+	envelopeFunc, ok := envelopeFuncs[mediaType]
+	if !ok {
+		return nil, &UnsupportedSignatureFormatError{MediaType: mediaType}
+	}
+	return envelopeFunc.newFunc(), nil
+}
+
+// ParseEnvelope generates an envelope by given envelope bytes with specified
+// media type.
+func ParseEnvelope(mediaType string, envelopeBytes []byte) (Envelope, error) {
+	envelopeFunc, ok := envelopeFuncs[mediaType]
+	if !ok {
+		return nil, &UnsupportedSignatureFormatError{MediaType: mediaType}
+	}
+	return envelopeFunc.parseFunc(envelopeBytes)
+}
--- a/signature/envelope.go
+++ b/signature/envelope.go
@@ -10,9 +10,12 @@
 // local crypto library or the external signing plugin.
 package signature
 
-import "fmt"
+import (
+	"fmt"
+	"sync"
+)
 
-// Envelope provides functions to basic functions to manipulate signatures.
+// Envelope provides basic functions to manipulate signatures.
 type Envelope interface {
 	// Sign generates and sign the envelope according to the sign request.
 	Sign(req *SignRequest) ([]byte, error)
@@ -41,7 +44,7 @@ type envelopeFunc struct {
 
 // envelopeFuncs maps envelope media type to corresponding constructors and
 // parsers.
-var envelopeFuncs map[string]envelopeFunc
+var envelopeFuncs sync.Map
 
 // RegisterEnvelopeType registers newFunc and parseFunc for the given mediaType.
 // Those functions are intended to be called when creating a new envelope.
@@ -50,14 +53,10 @@ func RegisterEnvelopeType(mediaType string, newFunc NewEnvelopeFunc, parseFunc P
 	if newFunc == nil || parseFunc == nil {
 		return fmt.Errorf("required functions not provided")
 	}
-	if envelopeFuncs == nil {
-		envelopeFuncs = make(map[string]envelopeFunc)
-	}
-
-	envelopeFuncs[mediaType] = envelopeFunc{
+	envelopeFuncs.Store(mediaType, envelopeFunc{
 		newFunc:   newFunc,
 		parseFunc: parseFunc,
-	}
+	})
 	return nil
 }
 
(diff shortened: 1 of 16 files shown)


gokarnm (changes requested):

signature/internal/base/envelope.go
```diff
@@ -0,0 +1,213 @@
...
+		return nil, err
+	}
+
+	return content, nil
+}
+
+// Content returns the validated signature information and payload.
+func (e *Envelope) Content() (*signature.EnvelopeContent, error) {
+	if len(e.Raw) == 0 {
+		return nil, &signature.MalformedSignatureError{Msg: "raw signature is empty"}
```
priteshbandi, line 85: ditto

signature/internal/base/envelope.go
```diff
@@ -0,0 +1,213 @@
...
+		return &signature.MalformedSignatureError{Msg: "expiry cannot be equal or before the signing time"}
+	}
+	return nil
+}
+
+// validatePayload performs validation of the payload.
+func validatePayload(payload *signature.Payload) error {
+	switch payload.ContentType {
+	case signature.MediaTypePayloadV1:
+		if len(payload.Content) == 0 {
```
priteshbandi, line 166: Should we move this check outside the the switch as payload cannot be of length 0?

signature/internal/base/envelope.go
```diff
@@ -0,0 +1,213 @@
...
+	if err := validatePayload(&req.Payload); err != nil {
+		return err
+	}
+
+	if err := validateSigningTime(req.SigningTime, req.Expiry); err != nil {
+		return err
+	}
+
+	if req.Signer == nil {
+		return &signature.MalformedSignatureError{Msg: "signer is nil"}
```
priteshbandi, line 111: Signature is not yet generated so instead of `MalformedSignatureError`, `MalformedSignRequestError` will be more appropriate.

signature/internal/base/envelope.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	_, err := req.Signer.KeySpec()
```
priteshbandi, line 114: We are missing validation on `SigningScheme`.

signature/internal/base/envelope.go
```diff
@@ -0,0 +1,213 @@
...
+	return validateCertificateChain(
+		info.CertificateChain,
+		signingTime,
+		info.SignatureAlgorithm,
+	)
+}
+
+// validateSigningTime checks that signing time is within the valid range of
+// time duration.
+func validateSigningTime(signingTime, expireTime time.Time) error {
```
priteshbandi, line 151: This functions validates that signing and expriy time are valid so we might want to rename it to `validateSigningAndExpiryTime`.

signature/internal/base/envelope.go
```diff
@@ -0,0 +1,213 @@
...
+func validateSignerInfo(info *signature.SignerInfo) error {
+	if len(info.Signature) == 0 {
+		return &signature.MalformedSignatureError{Msg: "signature not present or is empty"}
+	}
+
+	if info.SignatureAlgorithm == 0 {
+		return &signature.MalformedSignatureError{Msg: "SignatureAlgorithm is not present"}
+	}
+
+	signingTime := info.SignedAttributes.SigningTime
```
priteshbandi, line 137: We are missing validation for SigningScheme, SigningAgent.

signature/types.go
```diff
@@ -1,103 +1,149 @@
...
-	case SHA_256:
-		return crypto.SHA256
-	case SHA_384:
-		return crypto.SHA384
-	case SHA_512:
-		return crypto.SHA512
-	}
-	return 0
+// MediaTypePayloadV1 is the supported content type for signature's payload.
+const MediaTypePayloadV1 = "application/vnd.cncf.notary.payload.v1+json"
```
priteshbandi, line 27: Given that in future we can have multiple mediatype. Does this deserves its own type?

signature/types.go
```diff
@@ -1,103 +1,149 @@
...
+	// associated with the signing key used to generate the signature.
+	// The ordered list starts with the signing certificates, any intermediate
+	// certificates and ends with the root certificate.
+	CertificateChain []*x509.Certificate
+
+	// Signature is the bytes generated from the signature.
+	Signature []byte
+
+	// Payload represents payload in bytes and its content type.
+	Payload Payload
```
shizhMSFT, line 128: We need to remove it as it is duplicated.

signature/envelope.go
```diff
@@ -0,0 +1,92 @@
...
+// RegisterEnvelopeType registers newFunc and parseFunc for the given mediaType.
+// Those functions are intended to be called when creating a new envelope.
+// It will be called while inializing the built-in envelopes(JWS/COSE).
+func RegisterEnvelopeType(mediaType string, newFunc NewEnvelopeFunc, parseFunc ParseEnvelopeFunc) error {
+	if newFunc == nil || parseFunc == nil {
+		return fmt.Errorf("required functions not provided")
+	}
+	if envelopeFuncs == nil {
+		envelopeFuncs = make(map[string]envelopeFunc)
+	}
```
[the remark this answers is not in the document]
shizhMSFT: Let's document that this function is not go-routine safe.

signature/internal/base/envelope.go
shizhMSFT: @priteshbandi Can you point out the implementation of the validation before this PR so that they can be included.

priteshbandi: - [SigningScheme](https://github.com/notaryproject/notation-core-go/blob/main/signer/signer.go#L232-L234)
- Looks like there is no validation for SigningAgent

signature/types.go
```diff
@@ -1,103 +1,149 @@
...
+
+// EnvelopeContent represents a combination of payload to be signed and a parsed
+// signature envelope.
+type EnvelopeContent struct {
+	// SignerInfo is a parsed signature envelope.
+	SignerInfo SignerInfo
+
+	// Payload is payload to be signed.
+	Payload Payload
+}
```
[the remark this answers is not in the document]
gokarnm: We should remove Payload field from SignerInfo.

signature/errors.go
```diff
@@ -2,35 +2,44 @@ package signature
...
+	return fmt.Sprintf("signature is invalid. Error: %s", e.Err.Error())
+}
+
+// Unwrap unwraps the internal error.
+func (e *SignatureIntegrityError) Unwrap() error {
+	return e.Err
 }
 
 // MalformedSignatureError is used when Signature envelope is malformed.
 type MalformedSignatureError struct {
```
gokarnm, line 22:
```suggestion
type InvalidSignatureError struct {
```

signature/errors.go
```diff
@@ -40,56 +49,84 @@ func (e SignatureNotFoundError) Error() string {
...
-	if e.keyType != "" && e.keyLength != 0 {
-		return fmt.Sprintf("%q signing key of size %d is not supported", e.keyType, e.keyLength)
+	if e.Msg != "" {
+		return e.Msg
 	}
 	return "signing key is not supported"
 }
 
 // MalformedArgumentError is used when an argument to a function is malformed.
 type MalformedArgumentError struct {
```
gokarnm, line 75:
```suggestion
type InvalidArgumentError struct {
```

signature/errors.go
```diff
@@ -40,56 +49,84 @@ func (e SignatureNotFoundError) Error() string {
...
+	return fmt.Sprintf("%q param is malformed", e.Param)
+}
+
+// Unwrap returns the unwrapped error
+func (e *MalformedArgumentError) Unwrap() error {
+	return e.Err
 }
 
 // MalformedSignRequestError is used when SignRequest is malformed.
 type MalformedSignRequestError struct {
```
gokarnm, line 94:
```suggestion
type InvalidSignRequestError struct {
```

signature/errors.go
```diff
@@ -40,56 +49,84 @@ func (e SignatureNotFoundError) Error() string {
...
+
+// Error returns the default error message.
+func (e *SignatureEnvelopeNotFoundError) Error() string {
+	return "signature envelope is not present"
+}
+
+// EnvelopeKeyRepeatedError is used when repeated key name found in the envelope.
+type EnvelopeKeyRepeatedError struct {
+	Key string
 }
```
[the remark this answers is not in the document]
binbin-li: good call, I'll rename it.

signature/errors.go
```diff
@@ -40,56 +49,84 @@ func (e SignatureNotFoundError) Error() string {
...
 
-func (e SignatureAlgoNotSupportedError) Error() string {
-	return fmt.Sprintf("signature algorithm %q is not supported", e.alg)
+// Error returns the formatted error message.
+func (e *EnvelopeKeyRepeatedError) Error() string {
+	return fmt.Sprintf("repeated key: %q exists in the envelope.", e.Key)
 }
```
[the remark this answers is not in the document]
binbin-li: it will be used in following PRs.

signature/internal/base/envelope.go
```diff
@@ -0,0 +1,213 @@
...
+// Verify performs integrity and other signature specification related
+// validations.
+// It returns envelope content containing the payload to be signed and
+// SignerInfo object containing the information about the signature.
+//
+// Reference: https://github.com/notaryproject/notaryproject/blob/main/trust-store-trust-policy-specification.md#steps
+func (e *Envelope) Verify() (*signature.EnvelopeContent, error) {
+	// validation before the core verify process.
+	if len(e.Raw) == 0 {
+		return nil, &signature.MalformedSignatureError{}
```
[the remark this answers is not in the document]
binbin-li: That sounds reasonable to me.

signature/internal/base/envelope.go
binbin-li: As mentioned in community call, payload valiation will be moved to `notation-go`.

signature/envelope.go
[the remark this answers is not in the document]
priteshbandi: Or we can move initialization to variable declaration that should avoid multi threading issue.

signature/internal/base/envelope.go
priteshbandi: to summarize: remove any MediaType related checks in core go.

binbin-li: i see, so we keep the content size check in `notation-core-go`?

signature/internal/base/envelope.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	// core verify process.
+	content, err := e.Envelope.Verify()
```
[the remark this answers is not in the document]
binbin-li: In the current implementation, the exported Envelope and Internal Envelope shares the same `Verify()` api, so we don't pass it as argument.

signature/internal/base/envelope.go
binbin-li: thanks for calling it out!

signature/envelope.go
```diff
@@ -0,0 +1,92 @@
...
+// envelope, call RegisterEnvelopeType first during the initialization.
+//
+// A Signer is a structure used to sign payload generated after signature
+// envelope created. The underlying signing logic is provided by the underlying
+// local crypto library or the external signing plugin.
+package signature
+
+import "fmt"
+
+// Envelope provides functions to basic functions to manipulate signatures.
```
patrickzheng200, line 15: I'm not quite sure, what does 'functions to basic functions' mean here. Maybe "Envelope provides basic functions to manipulate signatures." ?

signature/signer.go
```diff
@@ -0,0 +1,150 @@
...
+// Signer is used to sign bytes generated after signature envelope created.
+type Signer interface {
+	// Sign signs the payload and returns the raw signature and certificates.
+	Sign(payload []byte) ([]byte, []*x509.Certificate, error)
+
+	// KeySpec returns the key specification.
+	KeySpec() (KeySpec, error)
+}
+
+// LocalSigner is used by built-in signers to sign only.
```
patrickzheng200, line 21: nit: // LocalSigner is only used by built-in signers to sign.

signature/envelope.go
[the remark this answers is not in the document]
binbin-li: thanks for calling it out, I'll use sync.Map to avoid thread-safe issues.

gokarnm (approved):
LGTM!

patrickzheng200 (approved):
LGTM

signature/internal/base/envelope.go
```diff
@@ -0,0 +1,221 @@
...
+	if err = validateEnvelopeContent(content); err != nil {
+		return nil, err
+	}
+
+	return content, nil
+}
+
+// validateSignRequest performs basic set of validations on SignRequest struct.
+func validateSignRequest(req *signature.SignRequest) error {
+	if err := validatePayload(&req.Payload); err != nil {
```
priteshbandi, line 102: This returns `InvalidSignatureError` which is not correct since signature is not yet generated.

I think we will need to verify this for all the validations done in `validateSignRequest`

signature/internal/base/envelope.go
```diff
@@ -0,0 +1,221 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// validate certificate chain
+	content, err := e.Envelope.Content()
+	if err != nil {
+		return nil, err
+	}
+	signerInfo := content.SignerInfo
```
shizhMSFT, line 42: nit: I know this is for readability but we really don't need to have a copy of `SignerInfo`.

signature/internal/base/envelope.go
priteshbandi: Yes, lets keep content length check because if content is empty than what are we going to sign.

signature/internal/base/envelope.go
binbin-li: Thanks for catching it. I have updated to return `InvalidSignRequestError` in `validateSignRequest` instead. But I'll keep this validation stay here.

shizhMSFT (approved):
LGTM

PR #75, opened 2022-09-14
feat: added COSE envelope implementation

patrickzheng200:
This PR adds COSE envelope implementation in cose/envelope.go.
This is based on the previous merged [refactor envelope and signer to support cose](https://github.com/notaryproject/notation-core-go/pull/73 ).

image


signature/cose/envelope.go
```diff
@@ -0,0 +1,595 @@
...
+	}
+	err = e.base.Verify(nil, verifier)
+	if err != nil {
+		return nil, &signature.SignatureIntegrityError{Err: err}
+	}
+
+	// extract content
+	content, err := e.Content()
+	if err != nil {
+		return nil, &signature.InvalidSignatureError{Msg: err.Error()}
```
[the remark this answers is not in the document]
patrickzheng200: Good catch. Will update.

signature/cose/envelope.go
```diff
@@ -0,0 +1,595 @@
...
+	}
+	signingScheme := signature.SigningScheme(signingSchemeString)
+	signerInfo.SignedAttributes.SigningScheme = signingScheme
+
+	// populate signerInfo.SignedAttributes.SigningTime
+	signingTimeLabel, ok := signingSchemeTimeLabelMap[signingScheme]
+	if !ok {
+		return &signature.InvalidSignatureError{Msg: "unsupported signingScheme: " + signingSchemeString}
+	}
+	signingTime, ok := protected[signingTimeLabel].(int64)
```
chloeyin, line 495: Do we need to check `signingTimeLabel` exists in the `protected` first?

signature/cose/envelope.go
```diff
@@ -0,0 +1,595 @@
...
+		}
+		extendedAttributeKeys = append(extendedAttributeKeys, label)
+	}
+
+	return extendedAttributeKeys, nil
+}
+
+// generateExtendedAttributes generates []signature.Attribute during SignerInfo process
+func generateExtendedAttributes(extendedAttributeKeys []string, protected cose.ProtectedHeader) ([]signature.Attribute, error) {
+	criticalHeaders, ok := protected[cose.HeaderLabelCritical].([]interface{})
```
chloeyin, line 571: Same. Do we need to check key exists first?

signature/cose/envelope.go
patrickzheng200: signingTime, ok := protected[signingTimeLabel].(int64) should check both existence and type assertion right?

signature/cose/envelope.go
patrickzheng200: > Same. Do we need to check key exists first?

Same here. this line should check both existence and type assertion?

signature/cose/envelope.go
chloeyin: > signingTime, ok := protected[signingTimeLabel].(int64) should check both existence and type assertion right?

Yes. Just we don't know which error it is. But if we don't need the detail message, it's ok 😄

signature/cose/envelope.go
```diff
@@ -0,0 +1,595 @@
...
+	// signerInfo fields
+	err := parseProtectedHeaders(e.base.Headers.Protected, &signerInfo)
+	if err != nil {
+		return nil, &signature.InvalidSignatureError{Msg: err.Error()}
+	}
+
+	// parse unprotected headers of COSE envelope
+	certs, ok := e.base.Headers.Unprotected[cose.HeaderLabelX5Chain].([]interface{})
+	if !ok || len(certs) == 0 {
+		return nil, &signature.InvalidSignatureError{Msg: "COSE envelope malformed certificate chain"}
```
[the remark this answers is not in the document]
patrickzheng200: > nit: malformed => invalid, to be consistent within the package.

We are actually having a problem here. The standard library [crypto/x509](https://cs.opensource.google/go/go/+/refs/tags/go1.19.1:src/crypto/x509/parser.go;drc=f78efc0178d51c02beff8a8203910dc0a9c6e953;l=789) uses term 'malformed' on certificates.

signature/cose/envelope.go
```diff
@@ -0,0 +1,595 @@
...
+	case signature.KeyTypeRSA:
+		switch keySpec.Size {
+		case 2048:
+			return cose.AlgorithmPS256, nil
+		case 3072:
+			return cose.AlgorithmPS384, nil
+		case 4096:
+			return cose.AlgorithmPS512, nil
+		default:
+			return 0, &signature.UnsupportedSigningKeyError{Msg: "RSA: key size not supported"}
```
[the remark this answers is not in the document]
patrickzheng200: Code updated.

go.mod
```diff
@@ -3,3 +3,9 @@ module github.com/notaryproject/notation-core-go
 go 1.18
 
 require github.com/golang-jwt/jwt/v4 v4.4.2
+
+require (
+	github.com/fxamacker/cbor/v2 v2.4.0 // indirect
+	github.com/veraison/go-cose v1.0.0-rc.1
```
shizhMSFT, line 9: Shouldn't we use `github.com/veraison/go-cose v1.0.0-rc.1.0.20220824135457-9d2fab636b83`?

go.mod
```diff
@@ -3,3 +3,9 @@ module github.com/notaryproject/notation-core-go
...
+	github.com/x448/float16 v0.8.4 // indirect
+)
```
shizhMSFT, line 11: We should have a `require` block of direct dependencies and a `require` block of indirect dependencies since go `1.17`.

signature/cose/envelope.go
```diff
@@ -0,0 +1,587 @@
...
+	"io"
+	"strconv"
+	"time"
+
+	"github.com/notaryproject/notation-core-go/signature"
+	"github.com/notaryproject/notation-core-go/signature/internal/base"
+	"github.com/veraison/go-cose"
+)
+
+// COSE signature envelope blob mediaType
```
shizhMSFT, line 17: The comment block should begin with `MediaTypeEnvelope ...` as per golang convension.

shizhMSFT (approved):
LGTM

signature/cose/envelope.go
```diff
@@ -0,0 +1,587 @@
...
+// Note: Verfiy only verifies integrity of the given COSE envelope
+func (e *envelope) Verify() (*signature.EnvelopeContent, error) {
+	// sanity check
+	if e.base == nil {
+		return nil, &signature.InvalidSignatureError{Msg: "missing COSE signature envelope"}
+	}
+
+	certs, ok := e.base.Headers.Unprotected[cose.HeaderLabelX5Chain].([]interface{})
+	if !ok || len(certs) == 0 {
+		return nil, &signature.InvalidSignatureError{Msg: "COSE envelope invalid certificate chain"}
```
priteshbandi, line 236: We should keep error message consistent with JWS.

https://github.com/notaryproject/notation-core-go/blob/main/signature/jws/envelope.go#L96
```suggestion
		return nil, &signature.InvalidSignatureError{Msg: "certificate chain is not present"}
```

signature/cose/envelope.go
```diff
@@ -0,0 +1,587 @@
...
+	}
+	certRaw, ok := certs[0].([]byte)
+	if !ok {
+		return nil, &signature.InvalidSignatureError{Msg: "COSE envelope malformed leaf certificate"}
+	}
+	cert, err := x509.ParseCertificate(certRaw)
+	if err != nil {
+		return nil, &signature.InvalidSignatureError{Msg: err.Error()}
```
priteshbandi, line 244:
```suggestion
		return nil, &signature.InvalidSignatureError{Msg: "malformed leaf certificate"}
```

signature/cose/envelope.go
```diff
@@ -0,0 +1,587 @@
...
+	}
+
+	// core verify process, verify integrity of COSE envelope
+	publicKeyAlg, err := getSignatureAlgorithm(cert)
+	if err != nil {
+		return nil, &signature.InvalidSignatureError{Msg: err.Error()}
+	}
+	verifier, err := cose.NewVerifier(publicKeyAlg, cert.PublicKey)
```
priteshbandi, line 252: We should add validation around supported algorithms? [veraison/go-cose](https://github.com/veraison/go-cose/blob/main/verifier.go) supports `AlgorithmEd25519` which notary doesnt and if go-cose might add support for symmetric keys that would become a major security issue for notation.

signature/cose/envelope.go
```diff
@@ -0,0 +1,587 @@
...
+}
+
+// Given a COSE envelope, extracts its signature.SignerInfo.
+func (e *envelope) signerInfo() (*signature.SignerInfo, error) {
+	var signerInfo signature.SignerInfo
+
+	// parse signature of COSE envelope, populate signerInfo.Signature
+	sig := e.base.Signature
+	if len(sig) == 0 {
+		return nil, &signature.InvalidSignatureError{Msg: "COSE envelope missing signature"}
```
priteshbandi, line 309:
```suggestion
		return nil, &signature.InvalidSignatureError{Msg: "signature missing in COSE envelope"}
```

signature/cose/envelope.go
```diff
@@ -0,0 +1,587 @@
...
+	err := parseProtectedHeaders(e.base.Headers.Protected, &signerInfo)
+	if err != nil {
+		return nil, &signature.InvalidSignatureError{Msg: err.Error()}
+	}
+
+	// parse unprotected headers of COSE envelope
+	certs, ok := e.base.Headers.Unprotected[cose.HeaderLabelX5Chain].([]interface{})
+	if !ok || len(certs) == 0 {
+		return nil, &signature.InvalidSignatureError{Msg: "COSE envelope invalid certificate chain"}
+	}
```
priteshbandi, line 324: Can be removed as its being already performed by https://github.com/notaryproject/notation-core-go/blob/main/signature/internal/base/envelope.go#L188-L190

signature/cose/envelope.go
```diff
@@ -0,0 +1,587 @@
...
+		certChain = append(certChain, cert)
+	}
+	// populate signerInfo.CertificateChain
+	signerInfo.CertificateChain = certChain
+
+	// populate signerInfo.UnsignedAttributes.SigningAgent
+	if h, ok := e.base.Headers.Unprotected[headerLabelSigningAgent].(string); ok {
+		signerInfo.UnsignedAttributes.SigningAgent = h
+	}
+
```
priteshbandi, line 344: missing TODO: For timestamp signature

signature/cose/envelope.go
```diff
@@ -0,0 +1,587 @@
...
+// Sign implements cose.Signer interface.
+func (signer *localSigner) Sign(rand io.Reader, payload []byte) ([]byte, error) {
+	coseSigner, err := cose.NewSigner(signer.alg, signer.cryptoSigner)
+	if err != nil {
+		return nil, err
+	}
+	return coseSigner.Sign(rand, payload)
+}
+
```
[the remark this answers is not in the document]
patrickzheng200: FYI, the above suggestion does not compile. But I get your point, will update accordingly.

signature/cose/envelope.go
```diff
@@ -0,0 +1,587 @@
...
+// validateCritHeaders does a two-way check, namely:
+// 1. validate that all critical headers are present in the protected bucket
+// 2. validate that all required headers(as per spec) are marked critical
+func validateCritHeaders(protected cose.ProtectedHeader) ([]string, error) {
+	// This ensures all critical headers are present in the protected bucket.
+	labels, err := protected.Critical()
+	if err != nil {
+		return nil, err
+	}
+
```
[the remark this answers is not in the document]
patrickzheng200: This case is already checked inside go-cose's Critical() function.

signature/cose/envelope.go
```diff
@@ -0,0 +1,587 @@
...
+	for label := range protected {
+		if contains(systemHeaders, label) {
+			continue
+		}
+		label, ok := label.(string)
+		if !ok {
+			return nil, &signature.InvalidSignatureError{Msg: "extendedAttributes key requires string type"}
+		}
+		extendedAttributeKeys = append(extendedAttributeKeys, label)
+	}
```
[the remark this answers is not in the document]
patrickzheng200:
```
// This ensures all critical headers are present in the protected bucket.
labels, err := protected.Critical()
```
This check includes extended attributes that are marked as critical. This procedure is done inside the go-cose library. 
As mentioned in the func doc:
// validateCritHeaders does a two-way check, namely:
// 1. validate that all critical headers are present in the protected bucket
// 2. validate that all required headers(as per spec) are marked critical

signature/cose/envelope.go
```diff
@@ -0,0 +1,587 @@
...
+	protected[cose.HeaderLabelCritical] = crit
+
+	return nil
+}
+
+// generateUnprotectedHeaders creates Unprotected Headers of the COSE envelope
+// during Sign process.
+func generateUnprotectedHeaders(req *signature.SignRequest, signer signer, unprotected cose.UnprotectedHeader) {
+	// signing agent
+	unprotected[headerLabelSigningAgent] = req.SigningAgent
```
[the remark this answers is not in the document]
patrickzheng200: Nice catch, updated.

signature/cose/envelope.go
```diff
@@ -0,0 +1,587 @@
...
+
+	// populate signerInfo.SignedAttributes.ExtendedAttributes
+	signerInfo.SignedAttributes.ExtendedAttributes, err = generateExtendedAttributes(extendedAttributeKeys, protected)
+	return err
+}
+
+// validateCritHeaders does a two-way check, namely:
+// 1. validate that all critical headers are present in the protected bucket
+// 2. validate that all required headers(as per spec) are marked critical
+func validateCritHeaders(protected cose.ProtectedHeader) ([]string, error) {
```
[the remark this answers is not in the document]
patrickzheng200: Sure thing.

signature/cose/envelope.go
```diff
@@ -0,0 +1,587 @@
...
+
+	// critical headers
+	protected[cose.HeaderLabelCritical] = crit
+
+	return nil
+}
+
+// generateUnprotectedHeaders creates Unprotected Headers of the COSE envelope
+// during Sign process.
+func generateUnprotectedHeaders(req *signature.SignRequest, signer signer, unprotected cose.UnprotectedHeader) {
```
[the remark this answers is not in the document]
patrickzheng200: The generateUnprotectedHeaders is intended to align with its counterpart, generateProtectedHeaders.

signature/cose/envelope.go
patrickzheng200: Yes, 	publicKeyAlg, err := getSignatureAlgorithm(cert) checks the algorithm, we only support RSA and EC.

signature/cose/envelope.go
patrickzheng200: This is an early break. The workflow is Content() -> signerInfo() -> validateEnvelopeContent() -> validateSignerInfo() -> validateCertificateChain(). The above code early breaks at the 2nd step, signerInfo().

signature/cose/envelope.go
```diff
@@ -439,7 +436,9 @@ func generateProtectedHeaders(req *signature.SignRequest, protected cose.Protect
 // during Sign process.
 func generateUnprotectedHeaders(req *signature.SignRequest, signer signer, unprotected cose.UnprotectedHeader) {
 	// signing agent
-	unprotected[headerLabelSigningAgent] = req.SigningAgent
+	if strings.TrimSpace(req.SigningAgent) != "" {
```
shizhMSFT, line 439: nit: we don't need to check space here since it is not a user-typed input.
```suggestion
	if req.SigningAgent != "" {
```

signature/cose/envelope.go
```diff
@@ -104,18 +105,18 @@ func (signer *remoteSigner) CertificateChain() []*x509.Certificate {
...
 }
 
-// localSigner implements signer interface.
-// It is used in Sign process when go-cose's built-in signer is desired.
 type localSigner struct {
-	base         signature.LocalSigner
-	alg          cose.Algorithm
-	cryptoSigner crypto.Signer
-	certs        []*x509.Certificate
+	coseSigner cose.Signer
```
shizhMSFT, line 109: How about
```suggestion
	cose.Signer
```
Then we don't even need to implement `Algorithm()` and `Sign(rand io.Reader, payload []byte)`.

patrickzheng200: Good idea! Updated the code.

signature/cose/envelope.go
```diff
@@ -130,23 +129,13 @@ func newLocalSigner(base signature.LocalSigner) (*localSigner, error) {
 			return nil, err
 		}
 		return &localSigner{
-			coseSigner: coseSigner,
-			certs:      certs,
+			coseSigner,
+			certs,
```
shizhMSFT, line 133: nit:
```suggestion
			Signer: coseSigner,
			certs:  certs,
```

shizhMSFT (approved):
LGTM with suggestions.

signature/cose/envelope.go
```diff
@@ -0,0 +1,576 @@
...
+)
+
+// Unprotected Headers
+// https://github.com/notaryproject/notaryproject/blob/cose-envelope/signature-envelope-cose.md
+const (
+	headerLabelTimeStampSignature = "io.cncf.notary.timestampSignature"
+	headerLabelSigningAgent       = "io.cncf.notary.signingAgent"
+)
+
+// Map of signature.Algorithm to cose.Algorithm
```
binbin-li, line 42: nit: looks like the description reverses the key and value?

signature/cose/envelope.go
```diff
@@ -0,0 +1,576 @@
...
+var coseAlgSignatureAlgMap = map[cose.Algorithm]signature.Algorithm{
+	cose.AlgorithmPS256: signature.AlgorithmPS256,
+	cose.AlgorithmPS384: signature.AlgorithmPS384,
+	cose.AlgorithmPS512: signature.AlgorithmPS512,
+	cose.AlgorithmES256: signature.AlgorithmES256,
+	cose.AlgorithmES384: signature.AlgorithmES384,
+	cose.AlgorithmES512: signature.AlgorithmES512,
+}
+
+// Map of signingScheme to signingTime header label
```
binbin-li, line 52: nit: missing period in the end. could you check other comments as well？

signature/cose/envelope.go
```diff
@@ -0,0 +1,576 @@
...
+		SignerInfo: *signerInfo,
+		Payload:    *payload,
+	}, nil
+}
+
+// Given a COSE envelope, extracts its signature.Payload.
+func (e *envelope) payload() (*signature.Payload, error) {
+	cty, ok := e.base.Headers.Protected[cose.HeaderLabelContentType]
+	if !ok {
+		return nil, &signature.InvalidSignatureError{Msg: "missing content type"}
```
binbin-li, line 274: I feel `invalid content type ` or `unsupported content type` might make more sense.

signature/cose/envelope.go
```diff
@@ -0,0 +1,576 @@
...
+		return nil, &signature.InvalidSignatureError{Msg: "invalid critical headers"}
+	}
+	var extendedAttr []signature.Attribute
+	for _, key := range extendedAttributeKeys {
+		extendedAttr = append(extendedAttr, signature.Attribute{
+			Key:      key,
+			Critical: contains(criticalHeaders, key),
+			Value:    protected[key],
+		})
+
```
binbin-li, line 563: nit: remove empty liine

signature/cose/envelope.go
patrickzheng200: Line 272 can only check if cose.HeaderLabelContentType exist, it does not validate it.

binbin-li (approved):
lgtm

shizhMSFT (approved):
LGTM

signature/cose/envelope.go
```diff
@@ -0,0 +1,575 @@
...
+	protected[cose.HeaderLabelCritical] = crit
+
+	return nil
+}
+
+// generateUnprotectedHeaders creates Unprotected Headers of the COSE envelope
+// during Sign process.
+func generateUnprotectedHeaders(req *signature.SignRequest, signer signer, unprotected cose.UnprotectedHeader) {
+	// signing agent
+	if req.SigningAgent != "" {
```
priteshbandi, line 428: We should add empty string `'   '` check.

priteshbandi (approved):
LGTM with one suggestion.

signature/cose/envelope.go
patrickzheng200: > We should add empty string `' '` check.

Shiwei had a different opinion on this one, https://github.com/notaryproject/notation-core-go/pull/75#discussion_r978410229. Since req.SigningAgent is not a user input, looks like we don't need to check empty here?

PR #87, opened 2022-10-06
Move signingTime verification to notation-go as TS/TP evaluation

priteshbandi:


signature/internal/base/envelope.go
```diff
@@ -156,7 +156,7 @@ func validateSignerInfo(info *signature.SignerInfo) error {
 
 	return validateCertificateChain(
 		info.CertificateChain,
-		signingTime,
+		time.Time{},
```
shizhMSFT, line 159: Passing a zero time is a security vulnerability. What if line 46 `content.SignerInfo.SignedAttributes.SigningTime` is set to zero by the adversary? It will bypass the check unexpectedly.

It is suggested to change
```go
func validateCertificateChain(certChain []*x509.Certificate, signTime time.Time, expectedAlg signature.Algorithm) error
```
and related functions to
```go
func validateCertificateChain(certChain []*x509.Certificate, signTime *time.Time, expectedAlg signature.Algorithm) error
```
marking `signTime` optional.

priteshbandi: We have explicit check to not allow zero signing time and its performed both during signing and verification, https://github.com/notaryproject/notation-core-go/blob/199ff3766849257ff3b452d6312b60e7df34b902/signature/internal/base/envelope.go#L167-L169

Also, `content.SignerInfo.SignedAttributes.SigningTime` is asigned parameter and if its modified, the signature will become invalid. Is there any specific security vulnerability you are thinking about?

patrickzheng200:
> Removing signingTime verification as its one by [notation-go](https://github.com/notaryproject/notation-go/blob/main/verification/verifier.go#L183-L187) during TS/TP evaluation

Looks like the code in [notation-go](https://github.com/notaryproject/notation-go/blob/main/verification/verifier.go#L183-L187) mentioned above does not check the SigningTime? It's for "verify x509 trust store based authenticity". I might be missing it, could you point to the singingTime related logic in notation-go?

signature/internal/base/envelope.go
patrickzheng200, line 159: Looks like the code you mentioned in [notation-go](https://github.com/notaryproject/notation-go/blob/main/verification/verifier.go#L183-L187) does not check the SigningTime? It's for "verify x509 trust store based authenticity". I might be missing it, could you point to the singingTime related logic in notation-go?

priteshbandi: [Here](https://github.com/notaryproject/notation-go/blob/2bcfd343f974938e8eeb76e233bc2526e0c36061/verification/verifier_helpers.go#L156-L170) we are performing verification.

patrickzheng200 (commented):
LGTM with other comments resolved.

signature/internal/base/envelope.go
shizhMSFT: I have concern that we will have regression in the future eventually.

jondonas (approved):

shizhMSFT (approved):
LGTM

PR #95, opened 2022-11-18
update: enabled cose to verify time in Tag1 Datetime format

patrickzheng200:
This PR enables COSE to verify time in Tag1 Datetime format.

Sign (Encode/MarshalCBOR):
in go-cose, time.Time will be encoded as numerical (in integer) representation of seconds since January 1, 1970 UTC.
For more details: https://github.com/fxamacker/cbor/blob/7704fa5efaf3ef4ac35aff38f50f6ff567793072/encode.go#L75
                            https://github.com/veraison/go-cose/blob/f72b6bc06b90d1e9c07ab11332df92c7838e3027/cbor.go#L30

Verify (Decode/UnmarshalCBOR):
1. CBOR times (tag 0 and 1) decode to time.Time. (This applies to COSE signatures provided by PRSS/ESRP.)
2. decode CBOR uint/int to Go int64.  (This applies to signatures signed by Notation Sign. Although, user passes time.Time into signRequest in Notation Sign, after MarshalCBOR (during Sign) and UnmarshalCBOR (during Verify), this time.Time is actually converted to an int64 by go-cose.)
For more details: https://github.com/fxamacker/cbor/blob/7704fa5efaf3ef4ac35aff38f50f6ff567793072/decode.go#L52
                            https://github.com/veraison/go-cose/blob/f72b6bc06b90d1e9c07ab11332df92c7838e3027/cbor.go#L43

### DIFF
--- a/signature/cose/envelope.go
+++ b/signature/cose/envelope.go
@@ -4,6 +4,7 @@ import (
 	"crypto"
 	"crypto/rand"
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"io"
 	"strconv"
@@ -393,7 +394,11 @@ func generateProtectedHeaders(req *signature.SignRequest, protected cose.Protect
 	if !ok {
 		return &signature.InvalidSignRequestError{Msg: "signing scheme: require notary.x509 or notary.x509.signingAuthority"}
 	}
-	protected[signingTimeLabel] = req.SigningTime.Unix()
+	// in go-cose, time.Time will be encoded as numerical (in integer)
+	// representation of seconds since January 1, 1970 UTC.
+	// For more details: https://github.com/fxamacker/cbor/blob/7704fa5efaf3ef4ac35aff38f50f6ff567793072/encode.go#L75
+	// https://github.com/veraison/go-cose/blob/f72b6bc06b90d1e9c07ab11332df92c7838e3027/cbor.go#L30
+	protected[signingTimeLabel] = req.SigningTime
 	if signingTimeLabel == headerLabelAuthenticSigningTime {
 		crit = append(crit, headerLabelAuthenticSigningTime)
 	}
@@ -401,7 +406,7 @@ func generateProtectedHeaders(req *signature.SignRequest, protected cose.Protect
 	// expiry
 	if !req.Expiry.IsZero() {
 		crit = append(crit, headerLabelExpiry)
-		protected[headerLabelExpiry] = req.Expiry.Unix()
+		protected[headerLabelExpiry] = req.Expiry
 	}
 
 	// extended attributes
@@ -471,20 +476,18 @@ func parseProtectedHeaders(protected cose.ProtectedHeader, signerInfo *signature
 	if !ok {
 		return &signature.InvalidSignatureError{Msg: "unsupported signingScheme: " + signingSchemeString}
 	}
-	signingTime, ok := protected[signingTimeLabel].(int64)
-	if !ok {
+	signingTime, err := parseTime(protected[signingTimeLabel])
+	if err != nil {
 		return &signature.InvalidSignatureError{Msg: "invalid signingTime under signing scheme: " + signingSchemeString}
 	}
-	signerInfo.SignedAttributes.SigningTime = time.Unix(signingTime, 0)
+	signerInfo.SignedAttributes.SigningTime = signingTime
 
 	// populate signerInfo.SignedAttributes.Expiry
-	if exp, ok := protected[headerLabelExpiry]; ok {
-		expiry, ok := exp.(int64)
-		if !ok {
-			return &signature.InvalidSignatureError{Msg: "expiry requires int64 type"}
-		}
-		signerInfo.SignedAttributes.Expiry = time.Unix(expiry, 0)
+	expiry, err := parseTime(protected[headerLabelExpiry])
+	if err != nil {
+		return &signature.InvalidSignatureError{Msg: "invalid expiry"}
 	}
+	signerInfo.SignedAttributes.Expiry = expiry
 
 	// populate signerInfo.SignedAttributes.ExtendedAttributes
 	signerInfo.SignedAttributes.ExtendedAttributes, err = generateExtendedAttributes(extendedAttributeKeys, protected)
@@ -568,4 +571,22 @@ func contains(s []interface{}, e interface{}) bool {
 		}
 	}
 	return false
+}
+
+// parseTime parses time values from cose.ProtectedHeader
+// in go-cose,
+//  1. CBOR times (tag 0 and 1) decode to time.Time.
+//  2. decode CBOR uint/int to Go int64
+//
+// For more details: https://github.com/fxamacker/cbor/blob/7704fa5efaf3ef4ac35aff38f50f6ff567793072/decode.go#L52
+// https://github.com/veraison/go-cose/blob/f72b6bc06b90d1e9c07ab11332df92c7838e3027/cbor.go#L43
+func parseTime(timeValue interface{}) (time.Time, error) {
+	switch t := timeValue.(type) {
+	case time.Time:
+		return t, nil
+	case int64:
+		return time.Unix(t, 0), nil
+	default:
+		return time.Time{}, errors.New("invalid type of timeValue")
+	}
 }
--- a/signature/cose/envelope_test.go
+++ b/signature/cose/envelope_test.go
@@ -558,7 +558,7 @@ func TestSignerInfoErrors(t *testing.T) {
 		}
 		env.base.Headers.Protected[headerLabelExpiry] = "invalid"
 		_, err = env.Content()
-		expected := errors.New("expiry requires int64 type")
+		expected := errors.New("invalid expiry")
 		if !isErrEqual(expected, err) {
 			t.Fatalf("Content() expects error: %v, but got: %v.", expected, err)
 		}
--- a/signature/internal/base/envelope.go
+++ b/signature/internal/base/envelope.go
@@ -114,7 +114,7 @@ func validateSignRequest(req *signature.SignRequest) error {
 	if _, err := req.Signer.KeySpec(); err != nil {
 		return err
 	}
-	
+
 	return validateSigningSchema(req.SigningScheme)
 }
 


signature/cose/envelope.go
```diff
@@ -569,3 +572,21 @@ func contains(s []interface{}, e interface{}) bool {
...
+//  2. decode CBOR uint/int to Go int64
+//
+// For more details: https://github.com/fxamacker/cbor/blob/7704fa5efaf3ef4ac35aff38f50f6ff567793072/decode.go#L52
+// https://github.com/veraison/go-cose/blob/f72b6bc06b90d1e9c07ab11332df92c7838e3027/cbor.go#L43
+func parseTime(timeValue interface{}) (time.Time, error) {
+	switch t := timeValue.(type) {
+	case time.Time:
+		return t, nil
+	case int64:
+		return time.Unix(t, 0), nil
```
shizhMSFT, line 588: Per spec, we don't support integers without a tag, do we?

signature/cose/envelope.go
```diff
@@ -393,15 +394,19 @@ func generateProtectedHeaders(req *signature.SignRequest, protected cose.Protect
...
+	protected[signingTimeLabel] = req.SigningTime
 	if signingTimeLabel == headerLabelAuthenticSigningTime {
 		crit = append(crit, headerLabelAuthenticSigningTime)
 	}
 
 	// expiry
 	if !req.Expiry.IsZero() {
 		crit = append(crit, headerLabelExpiry)
-		protected[headerLabelExpiry] = req.Expiry.Unix()
+		protected[headerLabelExpiry] = req.Expiry
```
shizhMSFT, line 409: What's the resulted CBOR object of marshalling a `time.Time`? Do we have a unit test to cover it?

signature/cose/envelope.go
```diff
@@ -569,3 +572,21 @@ func contains(s []interface{}, e interface{}) bool {
...
+// For more details: https://github.com/fxamacker/cbor/blob/7704fa5efaf3ef4ac35aff38f50f6ff567793072/decode.go#L52
+// https://github.com/veraison/go-cose/blob/f72b6bc06b90d1e9c07ab11332df92c7838e3027/cbor.go#L43
+func parseTime(timeValue interface{}) (time.Time, error) {
+	switch t := timeValue.(type) {
+	case time.Time:
+		return t, nil
+	case int64:
+		return time.Unix(t, 0), nil
+	default:
+		return time.Time{}, errors.New("invalid type of timeValue")
```
shizhMSFT, line 590: nit:
```suggestion
		return time.Time{}, errors.New("invalid date/time type")
```

signature/cose/envelope.go
patrickzheng200: > Per spec, we don't support integers without a tag, do we?

User input a time.Time into signRequest -> Sign -> go-cose Marshalls/Encodes this time.Time into an integer
-> go-cose Unmarshalls/Decodes this integer to int64 -> Verify. Therefore, during Verify, we still need to check the int64 type.

signature/cose/envelope.go
patrickzheng200: > What's the resulted CBOR object of marshalling a `time.Time`? Do we have a unit test to cover it?

As mentioned above, CBOR marshals time.Time to integer. Added unit test to cover this.

patrickzheng200: Code updated per discussion.

signature/cose/envelope.go
patrickzheng200: Code updated per discussion.

signature/cose/envelope.go
```diff
@@ -331,6 +346,65 @@ func (e *envelope) signerInfo() (*signature.SignerInfo, error) {
...
+	// populate signerInfo.SignedAttributes.Expiry
+	if !e.isSign {
+		expiry, err := parseTime(protected[headerLabelExpiry])
+		if err != nil {
+			return &signature.InvalidSignatureError{Msg: "invalid expiry"}
+		}
+		signerInfo.SignedAttributes.Expiry = expiry
+	} else {
+		signerInfo.SignedAttributes.Expiry = e.expiry
+	}
```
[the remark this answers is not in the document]
priteshbandi: Might be an novice comment: can we use TimeMode functionality provided by cbor https://github.com/fxamacker/cbor/blob/7704fa5efaf3ef4ac35aff38f50f6ff567793072/encode.go#L225 ? Not sure if its accessible from COSE.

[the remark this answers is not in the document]
patrickzheng200: > Might be an novice comment: can we use TimeMode functionality provided by cbor https://github.com/fxamacker/cbor/blob/7704fa5efaf3ef4ac35aff38f50f6ff567793072/encode.go#L225 ? Not sure if its accessible from COSE.

Oh yeah, this is exactly what we are doing here. Besides TimeMode, we are also using the EncTagMode (https://github.com/fxamacker/cbor/blob/1c59246e5a38253ebfe278d3f809f9b89a343e47/tag.go#L96) to create the Tag1 Datetime object in Sign.

signature/cose/envelope.go
```diff
@@ -142,7 +150,10 @@ func (signer *localSigner) CertificateChain() []*x509.Certificate {
 }
 
 type envelope struct {
-	base *cose.Sign1Message
+	isSign      bool
```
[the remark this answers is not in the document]
patrickzheng200: > Why do we need a flag to indicate if its a sign or verify call? Can we use the fact that sign and veify are different methods?

Code updated. But we still need to check if it is a sign or verify. This is because in base.Envelope.Sign(), we are also calling e.Envelope.Content(). However, due to Encode/Marshal, cose.Headers.Protected[SigningTime] is a cbor.RawMessage. Therefore, we need to Decode it back to time.Time during Sign. 
We don't need the Decode in verify though. Because when the upstream caller uses ParseEnvelope, Unmarshal is called (https://github.com/notaryproject/notation-core-go/blob/cd56ef60fb85f67d17bc59ff96637a93395c99a1/signature/cose/envelope.go#L158). go-cose will do the Decode for us.

JeyJeyGao (approved):
LGTM

rgnote (approved):
LGTM

signature/cose/envelope.go
```diff
@@ -388,20 +404,34 @@ func generateProtectedHeaders(req *signature.SignRequest, protected cose.Protect
 	crit := []interface{}{headerLabelSigningScheme}
 	protected[headerLabelSigningScheme] = string(req.SigningScheme)
 
+	// generate Tag1 Datetime CBOR object
+	encMode, err := encOpts.EncMode()
+	if err != nil {
+		return &signature.InvalidSignRequestError{Msg: err.Error()}
+	}
```
shizhMSFT, line 411: `encMode` should be a global variable initialized in `init()`.

signature/cose/envelope.go
```diff
@@ -388,20 +404,34 @@ func generateProtectedHeaders(req *signature.SignRequest, protected cose.Protect
...
+
 	// signingTime/authenticSigningTime
 	signingTimeLabel, ok := signingSchemeTimeLabelMap[req.SigningScheme]
 	if !ok {
 		return &signature.InvalidSignRequestError{Msg: "signing scheme: require notary.x509 or notary.x509.signingAuthority"}
 	}
-	protected[signingTimeLabel] = req.SigningTime.Unix()
+	signingTimeCBOR, err := encMode.Marshal(req.SigningTime)
```
shizhMSFT, line 418: It deserves a function
```go
function encodeTime(t time.Time) (cbor.RawMessage, error)
```

signature/cose/envelope.go
```diff
@@ -466,25 +496,47 @@ func parseProtectedHeaders(protected cose.ProtectedHeader, signerInfo *signature
 	signingScheme := signature.SigningScheme(signingSchemeString)
 	signerInfo.SignedAttributes.SigningScheme = signingScheme
 
+	// decode Tag1 Datetime CBOR object into time.Time
+	decMode, err := decOpts.DecMode()
+	if err != nil {
+		return &signature.InvalidSignRequestError{Msg: err.Error()}
+	}
```
shizhMSFT, line 503: `decMode` should be a global variable initialized in `init()`.

signature/cose/envelope.go
```diff
@@ -466,25 +496,47 @@ func parseProtectedHeaders(protected cose.ProtectedHeader, signerInfo *signature
...
 	// populate signerInfo.SignedAttributes.SigningTime
 	signingTimeLabel, ok := signingSchemeTimeLabelMap[signingScheme]
 	if !ok {
 		return &signature.InvalidSignatureError{Msg: "unsupported signingScheme: " + signingSchemeString}
 	}
-	signingTime, ok := protected[signingTimeLabel].(int64)
-	if !ok {
-		return &signature.InvalidSignatureError{Msg: "invalid signingTime under signing scheme: " + signingSchemeString}
+	var signingTime time.Time
+	if isSign {
```
shizhMSFT, line 511: The `isSign` is not necessary here since `cbor.RawMessage` type will never be unmarshalled to `interface{}`.

shizhMSFT: Therefore, a `switch` statement will be sufficient. Then we can move all related logic into `parseTime()`.

signature/cose/envelope.go
```diff
@@ -17,10 +19,28 @@ import (
 // MediaTypeEnvelope is the COSE signature envelope blob mediaType.
 const MediaTypeEnvelope = "application/cose"
 
+var (
+	// encOpts is the encoding options used in Sign
+	encOpts cbor.EncOptions
+
+	// decOpts is the decoding options used in Content
+	decOpts cbor.DecOptions
+)
```
shizhMSFT, line 28: Instead of `encOpts` and `decOpts`, should we have `encMode` and `decMode` like in [go-cose](https://github.com/veraison/go-cose/blob/main/cbor.go)?

signature/cose/envelope.go
```diff
@@ -471,20 +500,18 @@ func parseProtectedHeaders(protected cose.ProtectedHeader, signerInfo *signature
...
-		expiry, ok := exp.(int64)
-		if !ok {
-			return &signature.InvalidSignatureError{Msg: "expiry requires int64 type"}
-		}
-		signerInfo.SignedAttributes.Expiry = time.Unix(expiry, 0)
+	expiry, err := parseTime(protected[headerLabelExpiry])
+	if err != nil {
+		return &signature.InvalidSignatureError{Msg: fmt.Sprintf("invalid expiry: %v", err)}
 	}
+	signerInfo.SignedAttributes.Expiry = expiry
```
shizhMSFT, line 514: The original `if` statement is missing.

signature/cose/envelope.go
```diff
@@ -569,3 +596,55 @@ func contains(s []interface{}, e interface{}) bool {
...
+// For more details: https://github.com/fxamacker/cbor/blob/7704fa5efaf3ef4ac35aff38f50f6ff567793072/decode.go#L52
+func parseTime(timeValue interface{}) (time.Time, error) {
+	switch t := timeValue.(type) {
+	case cbor.RawMessage:
+		decTime, err := decodeTime(t)
+		if err != nil {
+			return time.Time{}, err
+		}
+		return decTime, nil
+	case time.Time:
```
shizhMSFT, line 645: There is issue here about the `time.Time`. How do we ensure it is parsed from tag `1` instead of tag `0`?

patrickzheng200: created an issue to track this.

signature/cose/envelope.go
```diff
@@ -17,10 +19,38 @@ import (
...
+	// decOpts is the decoding mode used in Content
+	decMode cbor.DecMode
+)
+
 func init() {
+	var err error
+
 	if err := signature.RegisterEnvelopeType(MediaTypeEnvelope, NewEnvelope, ParseEnvelope); err != nil {
 		panic(err)
 	}
```
shizhMSFT, line 35: nit:
```suggestion
	err := signature.RegisterEnvelopeType(MediaTypeEnvelope, NewEnvelope, ParseEnvelope)
	if err != nil {
		panic(err)
	}
```

signature/cose/envelope.go
```diff
@@ -17,10 +19,38 @@ import (
 // MediaTypeEnvelope is the COSE signature envelope blob mediaType.
 const MediaTypeEnvelope = "application/cose"
 
+var (
+	// encMode is the encoding mode used in Sign
+	encMode cbor.EncMode
+
+	// decOpts is the decoding mode used in Content
```
shizhMSFT, line 26:
```suggestion
	// decMode is the decoding mode used in Content
```

signature/cose/envelope.go
```diff
@@ -569,3 +608,48 @@ func contains(s []interface{}, e interface{}) bool {
...
+//
+// For more details: https://github.com/fxamacker/cbor/blob/7704fa5efaf3ef4ac35aff38f50f6ff567793072/decode.go#L52
+func parseTime(timeValue interface{}) (time.Time, error) {
+	switch t := timeValue.(type) {
+	case cbor.RawMessage:
+		decTime, err := decodeTime(t)
+		if err != nil {
+			return time.Time{}, err
+		}
+		return decTime, nil
```
shizhMSFT, line 646: nit:
```suggestion
		return decodeTime(t)
```

shizhMSFT (approved):
LGTM

PR #98, opened 2022-11-25
update: validate Tag1 Datetime CBOR object

### LINKED ISSUE #97: Tell apart Tag0 and Tag1 Datetime CBOR object during COSE verification
This issue tracks COSE verification with Tag0 or Tag1 Datetime. In our specs we only accept Tag1 Datetime CBOR object during verification. We need a way to check the tag of datetime inside the signature and fail if it's a Tag0.

patrickzheng200:
This PR's purpose is to fix #97. Based on our [spec](https://github.com/notaryproject/notaryproject/blob/main/specs/signature-envelope-cose.md#protected-header), we will only support Tag1 Datetime in notary v2 COSE.

This PR adds validation of signing time and expiry (if exists) to make sure they are Tag1 Datetime CBOR objects. Objects other than Tag1 including Tag0 Datetime CBOR would trigger an error. 


signature/cose/envelope.go
```diff
@@ -497,18 +496,22 @@ func parseProtectedHeaders(protected cose.ProtectedHeader, signerInfo *signature
...
+	signingSchemeString := protected[headerLabelSigningScheme].(string)
 	signingScheme := signature.SigningScheme(signingSchemeString)
-	signerInfo.SignedAttributes.SigningScheme = signingScheme
-
-	// populate signerInfo.SignedAttributes.SigningTime
 	signingTimeLabel, ok := signingSchemeTimeLabelMap[signingScheme]
 	if !ok {
 		return &signature.InvalidSignatureError{Msg: "unsupported signingScheme: " + signingSchemeString}
 	}
+	signerInfo.SignedAttributes.SigningScheme = signingScheme
```
priteshbandi, line 506: Can we move this like to 503 for better readability?

signature/cose/envelope.go
```diff
@@ -639,12 +641,55 @@ func parseTime(timeValue interface{}) (time.Time, error) {
...
+	if len(rawProtected) == 0 {
+		return nil
+	}
+
+	var decoded []byte
+	err := decMode.Unmarshal(rawProtected, &decoded)
+	if err != nil {
+		return err
+	}
+	cborMap := make(map[interface{}]cbor.RawMessage)
```
priteshbandi, line 663:
```suggestion
	var cborMap map[interface{}]cbor.RawMessage
```

signature/cose/envelope.go
```diff
@@ -639,12 +641,55 @@ func parseTime(timeValue interface{}) (time.Time, error) {
...
+func validateTag(cborMap map[interface{}]cbor.RawMessage, label interface{}) error {
+	rawMsg, ok := cborMap[label]
+	if ok {
+		rawTag := &cbor.RawTag{}
+		err := rawTag.UnmarshalCBOR([]byte(rawMsg))
+		if err != nil {
+			return err
+		}
+		if rawTag.Number != 1 {
+			return errors.New("only supports Tag1 Datetime CBOR object")
```
priteshbandi, line 690:
```suggestion
			return errors.New("only Tag1 Datetime CBOR object is supported")
```

priteshbandi (approved):
Would rely on review from @shizhMSFT for cose related unmarshalling code.

signature/cose/envelope.go
```diff
@@ -639,12 +641,55 @@ func parseTime(timeValue interface{}) (time.Time, error) {
...
 	}
 
 	return time.Time{}, errors.New("invalid timeValue type")
 }
+
+// validateTimeTag checks RawProtected of the COSE envelope Headers. Returns
+// error if signing time and/or expiry is not a Tag1 Datetime CBOR object.
+func validateTimeTag(rawProtected cbor.RawMessage, signingTimeLabel string) error {
+	if len(rawProtected) == 0 {
+		return nil
```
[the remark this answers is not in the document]
patrickzheng200: > Throw an error if `rawProtected` is nil or empty?

Oh yeah, rawProtected is populated during decoding. But we would call Content() right after Sign as well. At that time, rawProtected is not populated yet. Hence, it's empty. 
Whereas during Verify(), we will UnmarshalCBOR() beforehand and populate the rawProtected field in this process. If we still get an empty rawProtected, an error will be output from go-cose.

signature/cose/envelope.go
```diff
@@ -497,18 +496,22 @@ func parseProtectedHeaders(protected cose.ProtectedHeader, signerInfo *signature
 	signerInfo.SignatureAlgorithm = sigAlg
 
 	// populate signerInfo.SignedAttributes.SigningScheme
-	signingSchemeString, ok := protected[headerLabelSigningScheme].(string)
-	if !ok {
-		return &signature.InvalidSignatureError{Msg: "invalid signingScheme"}
-	}
+	// headerLabelSigningScheme type already checked in validateCritHeaders
+	signingSchemeString := protected[headerLabelSigningScheme].(string)
```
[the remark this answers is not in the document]
shizhMSFT: Also, data can be accessed without verification.

signature/cose/envelope.go
```diff
@@ -583,10 +587,8 @@ func validateCritHeaders(protected cose.ProtectedHeader) ([]interface{}, error)
 // generateExtendedAttributes generates []signature.Attribute during
 // SignerInfo process.
 func generateExtendedAttributes(extendedAttributeKeys []interface{}, protected cose.ProtectedHeader) ([]signature.Attribute, error) {
-	criticalHeaders, ok := protected[cose.HeaderLabelCritical].([]interface{})
-	if !ok {
-		return nil, &signature.InvalidSignatureError{Msg: "invalid critical headers"}
-	}
+	// cose.HeaderLabelCritical type already checked in validateCritHeaders
+	criticalHeaders := protected[cose.HeaderLabelCritical].([]interface{})
```
shizhMSFT, line 591: same here

signature/cose/envelope.go
```diff
@@ -497,18 +496,22 @@ func parseProtectedHeaders(protected cose.ProtectedHeader, signerInfo *signature
...
 	signingTimeLabel, ok := signingSchemeTimeLabelMap[signingScheme]
 	if !ok {
 		return &signature.InvalidSignatureError{Msg: "unsupported signingScheme: " + signingSchemeString}
 	}
+
+	// validates Tag1 Datetime CBOR object for signing time and expiry
+	err = validateDateTimeTag(rawProtected, signingTimeLabel)
+	if err != nil {
+		return &signature.InvalidSignatureError{Msg: "validateTimeTag failed: " + err.Error()}
+	}
```
shizhMSFT, line 512: This code block is just for signing time. How about expiry?

signature/cose/envelope.go
```diff
@@ -497,27 +496,32 @@ func parseProtectedHeaders(protected cose.ProtectedHeader, signerInfo *signature
...
-
-	// populate signerInfo.SignedAttributes.SigningTime
 	signingTimeLabel, ok := signingSchemeTimeLabelMap[signingScheme]
 	if !ok {
 		return &signature.InvalidSignatureError{Msg: "unsupported signingScheme: " + signingSchemeString}
 	}
-	signingTime, err := parseTime(protected[signingTimeLabel])
+
+	// generate RawProtected CBOR map
+	cborMap, err := generateRawProtectedCBORMap(rawProtected)
```
shizhMSFT, line 510: `cborMap` is too generic. How about `headerMap` or `headers`?

Also, you might want to update the comment to be something like `parse CBOR map from raw protected header for tag validation`.

signature/cose/envelope.go
```diff
@@ -631,20 +635,61 @@ func decodeTime(timeRaw cbor.RawMessage) (time.Time, error) {
...
 	return time.Time{}, errors.New("invalid timeValue type")
 }
+
+// generateRawProtectedCBORMap checks RawProtected of the COSE envelope Headers.
+// Returns the unmarshalled CBOR map of the RawProtected header.
+func generateRawProtectedCBORMap(rawProtected cbor.RawMessage) (map[interface{}]cbor.RawMessage, error) {
+	var cborMap map[interface{}]cbor.RawMessage
+	if len(rawProtected) == 0 {
+		return cborMap, nil
+	}
```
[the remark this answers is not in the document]
shizhMSFT: nit: you may replace `interface{}` with `any`.

signature/cose/envelope.go
```diff
@@ -631,20 +635,61 @@ func decodeTime(timeRaw cbor.RawMessage) (time.Time, error) {
...
+
+	var decoded []byte
+	err := decMode.Unmarshal(rawProtected, &decoded)
+	if err != nil {
+		return cborMap, err
```
shizhMSFT, line 669: nit: code style
```suggestion
		return nil, err
```

signature/cose/envelope.go
```diff
@@ -631,20 +635,61 @@ func decodeTime(timeRaw cbor.RawMessage) (time.Time, error) {
...
-func parseTime(timeValue interface{}) (time.Time, error) {
-	switch t := timeValue.(type) {
+func parseTime(cborMap map[interface{}]cbor.RawMessage, label string, protected cose.ProtectedHeader) (time.Time, error) {
+	err := validateTag(cborMap, label)
+	if err != nil {
+		return time.Time{}, err
+	}
+
+	switch t := protected[label].(type) {
 	case cbor.RawMessage:
```
shizhMSFT, line 649: Can we add a comment block stating that `cbor.RawMessage` comes from the signing path?

signature/cose/envelope.go
```diff
@@ -631,20 +635,61 @@ func decodeTime(timeRaw cbor.RawMessage) (time.Time, error) {
...
 		return decodeTime(t)
-	// TODO: need a way to check the tag number of datetime inside the signature
-	// and fail if it's a Tag0. We only accept Tag1 datetime.
-	// https://github.com/notaryproject/notation-core-go/issues/97
 	case time.Time:
```
shizhMSFT, line 651: Can we add a comment block stating that `time.Time` comes from the verification path?

shizhMSFT: We only need to validate tag for verification path.

signature/cose/envelope.go
```diff
@@ -631,20 +635,61 @@ func decodeTime(timeRaw cbor.RawMessage) (time.Time, error) {
...
 // For more details: https://github.com/fxamacker/cbor/blob/7704fa5efaf3ef4ac35aff38f50f6ff567793072/decode.go#L52
-func parseTime(timeValue interface{}) (time.Time, error) {
-	switch t := timeValue.(type) {
+func parseTime(cborMap map[interface{}]cbor.RawMessage, label string, protected cose.ProtectedHeader) (time.Time, error) {
+	err := validateTag(cborMap, label)
+	if err != nil {
+		return time.Time{}, err
+	}
+
+	switch t := protected[label].(type) {
```
shizhMSFT, line 648: What if `t` is `nil`? Instead of `invalid timeValue type`, should we return something like `missing xxx header`?

signature/cose/envelope.go
```diff
@@ -631,20 +635,61 @@ func decodeTime(timeRaw cbor.RawMessage) (time.Time, error) {
...
+		if err != nil {
+			return err
+		}
+		if rawTag.Number != 1 {
+			return errors.New("only Tag1 Datetime CBOR object is supported")
+		}
+	}
+
+	return nil
+}
```
shizhMSFT, line 695: I suggest merging this function to `parseTime` since `ok` is always `true`.

signature/cose/envelope.go
```diff
@@ -631,20 +635,61 @@ func decodeTime(timeRaw cbor.RawMessage) (time.Time, error) {
...
+
+// validateTag validates the tag number of the input field. Returns error
+// if the tag number is not 1.
+func validateTag(cborMap map[interface{}]cbor.RawMessage, label interface{}) error {
+	rawMsg, ok := cborMap[label]
+	if ok {
+		rawTag := &cbor.RawTag{}
+		err := rawTag.UnmarshalCBOR([]byte(rawMsg))
+		if err != nil {
+			return err
```
shizhMSFT, line 687: Should we return a more verbose error message?

shizhMSFT: For example, time value without a tag.

signature/cose/envelope.go
patrickzheng200: Merged into parseTime. Added a check for `ok`, although theoretically, it should always be `true`.

signature/cose/envelope.go
patrickzheng200: Updated.

signature/cose/envelope.go
patrickzheng200: Updated.

signature/cose/envelope.go
patrickzheng200: Updated.

signature/cose/envelope.go
```diff
@@ -631,20 +637,54 @@ func decodeTime(timeRaw cbor.RawMessage) (time.Time, error) {
...
+		if !ok {
+			return time.Time{}, fmt.Errorf("headerMap is missing label %q", label)
+		}
+		rawTag := &cbor.RawTag{}
+		err := rawTag.UnmarshalCBOR([]byte(rawMsg))
+		if err != nil {
+			return time.Time{}, fmt.Errorf("header %q time value does not have a tag", label)
+		}
+		if rawTag.Number != 1 {
+			return time.Time{}, errors.New("only Tag1 Datetime CBOR object is supported")
```
shizhMSFT, line 660:
```suggestion
			return time.Time{}, errors.New("only Tag `1` Datetime CBOR object is supported")
```

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM

PR #101, opened 2023-01-07
timestamp: add fuzzer

AdamKorcz:
Adds a simple fuzzer for `FuzzParseSignedToken`.

Once merged, I will instruct OSS-Fuzz to run this fuzzer continuously.

Signed-off-by: AdamKorcz <adam@adalogics.com>

### DIFF
--- a/timestamp/fuzz_test.go
+++ b/timestamp/fuzz_test.go
@@ -1,0 +1,11 @@
+package timestamp
+
+import (
+	"testing"
+)
+
+func FuzzParseSignedToken(f *testing.F) {
+	f.Fuzz(func(t *testing.T, data []byte) {
+		_, _ = ParseSignedToken(data)
+	})
+}


timestamp/fuzz_test.go
```diff
@@ -0,0 +1,11 @@
+package timestamp
+
+import (
+	"testing"
+)
+
+func FuzzParseSignedToken(f *testing.F) {
+	f.Fuzz(func(t *testing.T, data []byte) {
+		_, _ = ParseSignedToken(data)
```
priteshbandi, line 9: Just an callout: As of now we are not performing timestamping so `ParseSignedToken` method is dead code. But we will be adding time-stamping support in future so it doesnt hurts to add a fuzzer.

shizhMSFT:
The `timestamp` package has been cleaned up by #102

AdamKorcz:
Closing since the target code has been removed.

PR #103, opened 2023-01-16
fix: move signaturetest to internal package

### LINKED ISSUE #104: signaturetest package should not be exported
Currently, the signature/signaturetest is an exported package. It should be moved to an internal folder to avoid using by public users.

JeyJeyGao:
the signaturetest package only used by signature unit test, so it should be moved to internal package.
Also added unit test to increase the code coverage.

Resolves #104 


signature/internal/signaturetest/algorithm.go
```diff
@@ -1,3 +1,5 @@
+// Package signaturetest includes variables and functions for signature
```
shizhMSFT, line 1: If it is for internal test purposes only, is it possible to name `algorithm.go` to `algorithm_test.go`, `signer.go` to `signer_test.go`?

JeyJeyGao: I tried it. However, it looks like variables and functions in a file endded with "_test.go" cannot be used out side the packge. Those functions are shared by signature/jws and signature/cose packages.

shizhMSFT: I see. Can we also write unit tests for those "test functions"? It will also impact the coverage.

JeyJeyGao: Added test cases.

patrickzheng200 (approved):
LGTM

priteshbandi (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #111, opened 2023-02-01
Adds a string conversion for algorithms and some formatting

byronchien:
String conversion for algorithm enum used for inspect. Other changes result from `gofmt`

suggested order for review: [notation-go #261](https://github.com/notaryproject/notation-go/pull/261) => [notation-core-go #111](https://github.com/notaryproject/notation-core-go/pull/111) (this one) => [notation #527](https://github.com/notaryproject/notation/pull/527) =>[notation #528](https://github.com/notaryproject/notation/pull/528)

Signed-off-by: Byron Chien <chienb@amazon.com>

### DIFF
--- a/signature/algorithm.go
+++ b/signature/algorithm.go
@@ -15,13 +15,26 @@ type Algorithm int
 //
 // Reference: https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#algorithm-selection
 const (
-	AlgorithmPS256 Algorithm = 1 + iota // RSASSA-PSS with SHA-256
-	AlgorithmPS384                      // RSASSA-PSS with SHA-384
-	AlgorithmPS512                      // RSASSA-PSS with SHA-512
-	AlgorithmES256                      // ECDSA on secp256r1 with SHA-256
-	AlgorithmES384                      // ECDSA on secp384r1 with SHA-384
-	AlgorithmES512                      // ECDSA on secp521r1 with SHA-512
+	UnsupportedAlgorithm Algorithm = iota
+	AlgorithmPS256                 = iota // RSASSA-PSS with SHA-256
+	AlgorithmPS384                        // RSASSA-PSS with SHA-384
+	AlgorithmPS512                        // RSASSA-PSS with SHA-512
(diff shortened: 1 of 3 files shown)


byronchien:
required for notation inspect (added in this PR: https://github.com/notaryproject/notation/pull/528)

priteshbandi:
Looks like there are couple of files with white space changes, can we drop those files from this PR?

signature/algorithm.go
```diff
@@ -15,14 +15,27 @@ type Algorithm int
...
+func (alg Algorithm) String() string {
+	return [...]string{
+		"UnsupportedAlgorithm",
+		"RSASSA-PSS-SHA-256",
+		"RSASSA-PSS-SHA-384",
+		"RSASSA-PSS-SHA-512",
+		"ECDSA-secp256r1-SHA-256",
+		"ECDSA-secp384r1-SHA-384",
+		"ECDSA-secp521r1-SHA-512",
+	}[alg]
```
patrickzheng200, line 36: Looks like the elements in this string array have to strictly follow the order defined in `const`.  
I've seen an alternative implementation from [go-cose](https://github.com/veraison/go-cose/blob/07c7d28e2c7eb9d6daa4619b98eceeaf7a8e5156/algorithm.go#L52). Using switch statements seems easier to maintain.

shizhMSFT (changes requested):
We don't have a spec on the standard string of those algorithms. AFAIK, those strings come from the plugin spec. Therefore, those changes should be in the `notation-go` repo.

byronchien:
> We don't have a spec on the standard string of those algorithms. AFAIK, those strings come from the plugin spec. Therefore, those changes should be in the `notation-go` repo.

ok, if [this helper](https://github.com/notaryproject/notation-go/blob/61ffb0982848f62509497eda9e0088c90368d7ff/plugin/proto/algorithm.go#L132) doesn't work, will add to notation-go instead. will verify before closing this PR

byronchien:
used the existing EncodeSigningAlgorithm helper, resolving

PR #120, opened 2023-02-09
fix: updated error messages for certificate chain validation in x509 package

### LINKED ISSUE #117: certificate chain validation error message is not returned
The error message for validate certificate is not returned. [Here is code](https://github.com/notaryproject/notation-core-go/blob/0cb6e7508c67ee9a3b3a2e680ae7841ec37fdd32/x509/cert_validations.go#L92).

https://github.com/notaryproject/notation-core-go/blob/0cb6e7508c67ee9a3b3a2e680ae7841ec37fdd32/x509/cert_validations.go#L91-L94

The `subject.CheckSignatureFrom` may return an error and the error message is not returned, which will finally cause an inaccurate error.

### LINKED ISSUE #119: error message may not right
In line 67 of the [code](https://github.com/notaryproject/notation-core-go/blob/0cb6e7508c67ee9a3b3a2e680ae7841ec37fdd32/x509/cert_validations.go#L67), if i == 0 which is the leaf cert, the error message still says that it is the intermediate certificate, so the error message may not right. It may need to return the CommonName of the cert to help use identify the invalid cert.

https://github.com/notaryproject/notation-core-go/blob/0cb6e7508c67ee9a3b3a2e680ae7841ec37fdd32/x509/cert_validations.go#L56-L78

### LINKED ISSUE #121: self-signed leaf cert error message not clear
https://github.com/notaryproject/notation-core-go/blob/0cb6e7508c67ee9a3b3a2e680ae7841ec37fdd32/x509/cert_validations.go#L43-L54
In line 50 or 53, if it return an error, it had better tell user that there is only one cert and the cert subject is xxxxx with error:xxxx
Also, it may be replaced by `isSelfSigned()` because it also need to make sure that the issuer and subject are the same.

patrickzheng200:
This PR would:
resolve #117
resolve #119
resolve #121 

### DIFF
--- a/x509/cert_validations.go
+++ b/x509/cert_validations.go
@@ -48,11 +48,13 @@ func validateCertChain(certChain []*x509.Certificate, expectedLeafEku x509.ExtKe
 		if signingTime != nil && (signingTime.Before(cert.NotBefore) || signingTime.After(cert.NotAfter)) {
 			return fmt.Errorf("certificate with subject %q was not valid at signing time of %s", cert.Subject, signingTime.UTC())
 		}
-
 		if err := cert.CheckSignature(cert.SignatureAlgorithm, cert.RawTBSCertificate, cert.Signature); err != nil {
-			return err
+			return fmt.Errorf("self-signed signing certificate with subject %q failed at checking signature. Error: %v", cert.Subject, err)
 		}
-		return validateLeafCertificate(cert, expectedLeafEku)
+		if err := validateLeafCertificate(cert, expectedLeafEku); err != nil {
+			return fmt.Errorf("self-signed signing certificate is invalid. Error: %v", err)
+		}
+		return nil
 	}
 
 	for i, cert := range certChain {
--- a/x509/cert_validations.go
+++ b/x509/cert_validations.go
@@ -49,10 +49,10 @@ func validateCertChain(certChain []*x509.Certificate, expectedLeafEku x509.ExtKe
 			return fmt.Errorf("certificate with subject %q was not valid at signing time of %s", cert.Subject, signingTime.UTC())
 		}
 		if err := cert.CheckSignature(cert.SignatureAlgorithm, cert.RawTBSCertificate, cert.Signature); err != nil {
-			return fmt.Errorf("self-signed signing certificate with subject %q failed at checking signature. Error: %v", cert.Subject, err)
+			return fmt.Errorf("invalid self-signed certificate. subject: %q. Error: %v", cert.Subject, err)
 		}
 		if err := validateLeafCertificate(cert, expectedLeafEku); err != nil {
-			return fmt.Errorf("self-signed signing certificate is invalid. Error: %v", err)
+			return fmt.Errorf("invalid self-signed certificate. Error: %v", err)
 		}
 		return nil
 	}
--- a/x509/cert_validations.go
+++ b/x509/cert_validations.go
@@ -49,10 +49,10 @@ func validateCertChain(certChain []*x509.Certificate, expectedLeafEku x509.ExtKe
 			return fmt.Errorf("certificate with subject %q was not valid at signing time of %s", cert.Subject, signingTime.UTC())
 		}
 		if err := cert.CheckSignature(cert.SignatureAlgorithm, cert.RawTBSCertificate, cert.Signature); err != nil {
-			return fmt.Errorf("invalid self-signed certificate. subject: %q. Error: %v", cert.Subject, err)
+			return fmt.Errorf("invalid self-signed certificate. subject: %q. Error: %w", cert.Subject, err)
 		}
 		if err := validateLeafCertificate(cert, expectedLeafEku); err != nil {
-			return fmt.Errorf("invalid self-signed certificate. Error: %v", err)
+			return fmt.Errorf("invalid self-signed certificate. Error: %w", err)
 		}
 		return nil
 	}
@@ -61,18 +61,23 @@ func validateCertChain(certChain []*x509.Certificate, expectedLeafEku x509.ExtKe
 		if signingTime != nil && (signingTime.Before(cert.NotBefore) || signingTime.After(cert.NotAfter)) {
 			return fmt.Errorf("certificate with subject %q was not valid at signing time of %s", cert.Subject, signingTime.UTC())
 		}
-		selfSigned, selfSignedError := isSelfSigned(cert)
 		if i == len(certChain)-1 {
+			selfSigned, selfSignedError := isSelfSigned(cert)
+			if selfSignedError != nil {
+				return fmt.Errorf("root certificate with subject %q is invalid or not self-signed. Certificate chain must end with a valid self-signed root certificate. Error: %v", cert.Subject, selfSignedError)
+			}
 			if !selfSigned {
-				if selfSignedError != nil {
-					return fmt.Errorf("root certificate with subject %q is invalid or not self-signed. Certificate chain must end with a valid self-signed root certificate. Error: %v", cert.Subject, selfSignedError)
-				}
 				return fmt.Errorf("root certificate with subject %q is not self-signed. Certificate chain must end with a valid self-signed root certificate", cert.Subject)
 			}
 		} else {
 			// This is to avoid extra/redundant multiple root cert at the end
 			// of certificate-chain
-			if selfSigned {
+			selfSigned, selfSignedError := isSelfSigned(cert)
+			// not checking selfSignedError != nil here because we expect
+			// a non-nil err. For a non-root certificate, it shouldn't be
+			// self-signed, hence CheckSignatureFrom would return a non-nil
+			// error.
+			if selfSignedError == nil && selfSigned {
 				if i == 0 {
 					return fmt.Errorf("leaf certificate with subject %q is self-signed. Certificate chain must not contain self-signed leaf certificate", cert.Subject)
 				}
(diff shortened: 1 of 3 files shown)


x509/cert_validations.go
```diff
@@ -57,16 +57,28 @@ func validateCertChain(certChain []*x509.Certificate, expectedLeafEku x509.ExtKe
 		if signingTime != nil && (signingTime.Before(cert.NotBefore) || signingTime.After(cert.NotAfter)) {
 			return fmt.Errorf("certificate with subject %q was not valid at signing time of %s", cert.Subject, signingTime.UTC())
 		}
-
+		selfSigned, selfSignedError := isSelfSigned(cert)
 		if i == len(certChain)-1 {
-			if !isSelfSigned(cert) {
-				return errors.New("certificate chain must end with a root certificate (root certificates are self-signed)")
+			if !selfSigned {
+				if selfSignedError != nil {
```
JeyJeyGao, line 63: if it is `selfSigned` but `selfSignedError`  is not nil, do we expect to skip the error?

patrickzheng200: > if it is `selfSigned` but `selfSignedError` is not nil, do we expect to skip the error?

This will never happen. Because when `isSelfSigned` returns true, `err` will always be `nil`: https://github.com/patrickzheng200/notation-core-go/blob/495fc78c6158d368c7a57a85d6094fe0dda5d036/x509/cert_validations.go#L107

x509/cert_validations.go
```diff
@@ -57,16 +57,28 @@ func validateCertChain(certChain []*x509.Certificate, expectedLeafEku x509.ExtKe
...
-			if isSelfSigned(cert) {
-				return errors.New("certificate chain must not contain self-signed intermediate certificates")
-			} else if nextCert := certChain[i+1]; !isIssuedBy(cert, nextCert) {
+			if selfSigned {
+				if i == 0 {
+					return fmt.Errorf("leaf certificate with subject %q is self-signed. Certificate chain must not contain self-signed leaf certificate", cert.Subject)
+				}
+				return fmt.Errorf("intermediate certificate with subject %q is self-signed. Certificate chain must not contain self-signed intermediate certificate", cert.Subject)
+			}
+			nextCert := certChain[i+1]
```
JeyJeyGao, line 76:
```suggestion
			parentCert := certChain[i+1]
```
nit

JeyJeyGao (approved):
LGTM with a suggestion

x509/cert_validations.go
```diff
@@ -46,28 +48,43 @@ func validateCertChain(certChain []*x509.Certificate, expectedLeafEku x509.ExtKe
...
-		return validateLeafCertificate(cert, expectedLeafEku)
+		return nil
 	}
 
 	for i, cert := range certChain {
 		if signingTime != nil && (signingTime.Before(cert.NotBefore) || signingTime.After(cert.NotAfter)) {
 			return fmt.Errorf("certificate with subject %q was not valid at signing time of %s", cert.Subject, signingTime.UTC())
 		}
-
+		selfSigned, selfSignedError := isSelfSigned(cert)
```
priteshbandi, line 64: Can this can be moved inside `if i == len(certChain)-1 {` ?

x509/cert_validations.go
```diff
@@ -46,28 +48,43 @@ func validateCertChain(certChain []*x509.Certificate, expectedLeafEku x509.ExtKe
 		if signingTime != nil && (signingTime.Before(cert.NotBefore) || signingTime.After(cert.NotAfter)) {
 			return fmt.Errorf("certificate with subject %q was not valid at signing time of %s", cert.Subject, signingTime.UTC())
 		}
-
 		if err := cert.CheckSignature(cert.SignatureAlgorithm, cert.RawTBSCertificate, cert.Signature); err != nil {
```
JeyJeyGao, line 51: Do we need to validate that it was issued by itself? `isSelfSigned()`

x509/cert_validations.go
```diff
@@ -46,28 +48,43 @@ func validateCertChain(certChain []*x509.Certificate, expectedLeafEku x509.ExtKe
...
-			return err
+			return fmt.Errorf("self-signed signing certificate with subject %q failed at checking signature. Error: %v", cert.Subject, err)
```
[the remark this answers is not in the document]
JeyJeyGao: what about Invalid self-signed leaf certificate

x509/cert_validations.go
patrickzheng200: Here's the func doc of [CheckSignature](https://pkg.go.dev/crypto/x509#Certificate.CheckSignature). The usage here is indeed checking if it is a self-signed certificate.
Just in case you would curious about why we are not using the `isSelfSigned` here. Since the self-signed cert is the only cert in the cert chain and it's very likely not issued by CA, this cert won't pass the `CheckSignatureFrom` check, which is used in our `isSelfSigned` func. (source code of [CheckSignatureFrom)](https://cs.opensource.google/go/go/+/refs/tags/go1.20:src/crypto/x509/x509.go;l=822).

x509/cert_validations.go
patrickzheng200: But it is also used under the `else` statement: https://github.com/notaryproject/notation-core-go/blob/0cb6e7508c67ee9a3b3a2e680ae7841ec37fdd32/x509/cert_validations.go#L67

x509/cert_validations.go
```diff
@@ -46,28 +48,43 @@ func validateCertChain(certChain []*x509.Certificate, expectedLeafEku x509.ExtKe
...
+		}
+		if err := validateLeafCertificate(cert, expectedLeafEku); err != nil {
+			return fmt.Errorf("self-signed signing certificate is invalid. Error: %v", err)
```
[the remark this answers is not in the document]
patrickzheng200: We don't need `subject` here. All `err` in this case already contain the `subject` field.

x509/cert_validations.go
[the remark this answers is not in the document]
patrickzheng200: > what about Invalid self-signed leaf certificate

Code at line 52 is about a cert chain with only a single cert (self-signed) in it. So I'm not very sure what do you mean by `leaf` here?

priteshbandi (approved):
LGTM

x509/cert_validations.go
```diff
@@ -46,28 +48,43 @@ func validateCertChain(certChain []*x509.Certificate, expectedLeafEku x509.ExtKe
 		if signingTime != nil && (signingTime.Before(cert.NotBefore) || signingTime.After(cert.NotAfter)) {
 			return fmt.Errorf("certificate with subject %q was not valid at signing time of %s", cert.Subject, signingTime.UTC())
 		}
-
 		if err := cert.CheckSignature(cert.SignatureAlgorithm, cert.RawTBSCertificate, cert.Signature); err != nil {
-			return err
+			return fmt.Errorf("invalid self-signed certificate. subject: %q. Error: %v", cert.Subject, err)
```
shizhMSFT, line 52: It is better to wrap the error.
```suggestion
			return fmt.Errorf("invalid self-signed certificate. subject: %q. Error: %w", cert.Subject, err)
```

x509/cert_validations.go
```diff
@@ -46,28 +48,43 @@ func validateCertChain(certChain []*x509.Certificate, expectedLeafEku x509.ExtKe
...
+		}
+		if err := validateLeafCertificate(cert, expectedLeafEku); err != nil {
+			return fmt.Errorf("invalid self-signed certificate. Error: %v", err)
```
shizhMSFT, line 55: It is better to wrap the error.
```suggestion
			return fmt.Errorf("invalid self-signed certificate. Error: %w", err)
```

x509/cert_validations.go
```diff
@@ -46,28 +48,43 @@ func validateCertChain(certChain []*x509.Certificate, expectedLeafEku x509.ExtKe
...
-		return validateLeafCertificate(cert, expectedLeafEku)
+		return nil
 	}
 
 	for i, cert := range certChain {
 		if signingTime != nil && (signingTime.Before(cert.NotBefore) || signingTime.After(cert.NotAfter)) {
 			return fmt.Errorf("certificate with subject %q was not valid at signing time of %s", cert.Subject, signingTime.UTC())
 		}
-
+		selfSigned, selfSignedError := isSelfSigned(cert)
```
shizhMSFT, line 64: Please check `selfSignedError` immediately after calling `isSelfSigned`, and never trust the value of `selfSigned` if `selfSignedError` is not `nil`.

x509/cert_validations.go
```diff
@@ -46,28 +48,43 @@ func validateCertChain(certChain []*x509.Certificate, expectedLeafEku x509.ExtKe
...
+			if selfSigned {
+				if i == 0 {
+					return fmt.Errorf("leaf certificate with subject %q is self-signed. Certificate chain must not contain self-signed leaf certificate", cert.Subject)
+				}
+				return fmt.Errorf("intermediate certificate with subject %q is self-signed. Certificate chain must not contain self-signed intermediate certificate", cert.Subject)
+			}
+			parentCert := certChain[i+1]
+			issuedBy, issuedByError := isIssuedBy(cert, parentCert)
+			if !issuedBy {
+				if issuedByError != nil {
```
shizhMSFT, line 84: Always check error and then use the return value.

shizhMSFT (approved):
LGTM

PR #129, opened 2023-03-13
fuzzing: run notation-core-go fuzzers in the CI

AdamKorcz:
This PR will run notation-co-go's fuzzers in the CI via OSS-Fuzz's CIFuzz. The fuzzers are available here: https://github.com/cncf/cncf-fuzzing/tree/main/projects/notary. The documentation on CIFuzz: https://google.github.io/oss-fuzz/getting-started/continuous-integration/

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,30 @@
+name: CIFuzz
+on:
+  pull_request:
+    paths-ignore:
+      - 'Documentation/**'
+permissions: read-all
+jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,30 @@
+name: CIFuzz
+on:
+  pull_request:
+    paths-ignore:
+      - 'Documentation/**'
```
shizhMSFT, line 5: Should it be
```suggestion
      - '.github/**'
```

PR #134, opened 2023-03-24
feat: adding OCSP revocation implementation

kody-kimberl:
This PR adds a new package that will perform OCSP revocation checking for a certificate chain.

Based on my design from #132 and the specification [here](https://github.com/notaryproject/notaryproject/blob/main/specs/trust-store-trust-policy.md#certificate-revocation-evaluation).

This PR addresses part of the following issue:
 - #124 

Once it is approved and merged, a second PR will implement this functionality into notation-go.

Signed-off-by: Kody Kimberl kody.kimberl.work@gmail.com

PR #141, opened 2023-04-21
fix: change OCSP hash and encoding

kody-kimberl:
This is a followup fix to #134 

After testing with real certificates, there are some discrepancies with what hashes are supported and the correct URL encoding.

---

The following do not support SHA256 hashes:
 - Microsoft
 - Entrust
 - Let's Encrypt
 - Digicert (sometimes)

As this represents a large percentage of public CAs, we should change the hashing algorithm to SHA1, which has been confirmed to be supported by all that were tested.

---

Additionally, the base64 URL encoding was not actually escaping URL characters, resulting in malformed requests. We need to change it to StdEncoding and query escape it.

Signed-off-by: Kody Kimberl kody.kimberl.work@gmail.com

### DIFF
--- a/revocation/ocsp/ocsp.go
+++ b/revocation/ocsp/ocsp.go
@@ -167,7 +167,7 @@ func extensionsToMap(extensions []pkix.Extension) map[string][]byte {
 func executeOCSPCheck(cert, issuer *x509.Certificate, server string, opts Options) (*ocsp.Response, error) {
 	// TODO: Look into other alternatives for specifying the Hash
 	// https://github.com/notaryproject/notation-core-go/issues/139
-	ocspRequest, err := ocsp.CreateRequest(cert, issuer, &ocsp.RequestOptions{Hash: crypto.SHA256})
+	ocspRequest, err := ocsp.CreateRequest(cert, issuer, &ocsp.RequestOptions{Hash: crypto.SHA1})
 	if err != nil {
 		return nil, GenericError{Err: err}
 	}
@@ -177,7 +177,7 @@ func executeOCSPCheck(cert, issuer *x509.Certificate, server string, opts Option
 		reader := bytes.NewReader(ocspRequest)
 		resp, err = opts.HTTPClient.Post(server, "application/ocsp-request", reader)
 	} else {
-		encodedReq := base64.URLEncoding.EncodeToString(ocspRequest)
+		encodedReq := url.QueryEscape(base64.StdEncoding.EncodeToString(ocspRequest))
 		var reqURL string
 		reqURL, err = url.JoinPath(server, encodedReq)
 		if err != nil {
--- a/revocation/ocsp/ocsp.go
+++ b/revocation/ocsp/ocsp.go
@@ -181,11 +181,16 @@ func executeOCSPCheck(cert, issuer *x509.Certificate, server string, opts Option
 	}
 
 	var resp *http.Response
-	if base64.StdEncoding.EncodedLen(len(ocspRequest)) >= 255 {
+	postRequired := base64.StdEncoding.EncodedLen(len(ocspRequest)) >= 255
+	var encodedReq string
+	if !postRequired {
+		encodedReq = url.QueryEscape(base64.StdEncoding.EncodeToString(ocspRequest))
+		postRequired = len(encodedReq) >= 255
+	}
+	if postRequired {
 		reader := bytes.NewReader(ocspRequest)
 		resp, err = opts.HTTPClient.Post(server, "application/ocsp-request", reader)
 	} else {
-		encodedReq := url.QueryEscape(base64.StdEncoding.EncodeToString(ocspRequest))
 		var reqURL string
 		reqURL, err = url.JoinPath(server, encodedReq)
 		if err != nil {
--- a/revocation/ocsp/ocsp.go
+++ b/revocation/ocsp/ocsp.go
@@ -185,18 +185,18 @@ func executeOCSPCheck(cert, issuer *x509.Certificate, server string, opts Option
 	var encodedReq string
 	if !postRequired {
 		encodedReq = url.QueryEscape(base64.StdEncoding.EncodeToString(ocspRequest))
-		postRequired = len(encodedReq) >= 255
-	}
-	if postRequired {
-		reader := bytes.NewReader(ocspRequest)
-		resp, err = opts.HTTPClient.Post(server, "application/ocsp-request", reader)
-	} else {
-		var reqURL string
-		reqURL, err = url.JoinPath(server, encodedReq)
-		if err != nil {
-			return nil, GenericError{Err: err}
+		if len(encodedReq) < 255 {
+			var reqURL string
+			reqURL, err = url.JoinPath(server, encodedReq)
+			if err != nil {
+				return nil, GenericError{Err: err}
+			}
+			resp, err = opts.HTTPClient.Get(reqURL)
+		} else {
+			resp, err = postRequest(ocspRequest, server, opts.HTTPClient)
 		}
-		resp, err = opts.HTTPClient.Get(reqURL)
+	} else {
+		resp, err = postRequest(ocspRequest, server, opts.HTTPClient)
 	}
 
 	if err != nil {
@@ -231,6 +231,11 @@ func executeOCSPCheck(cert, issuer *x509.Certificate, server string, opts Option
 	}
 
 	return ocsp.ParseResponseForCert(body, cert, issuer)
+}
+
+func postRequest(req []byte, server string, httpClient *http.Client) (*http.Response, error) {
+	reader := bytes.NewReader(req)
+	return httpClient.Post(server, "application/ocsp-request", reader)
 }
 
 func toServerResult(server string, err error) *result.ServerResult {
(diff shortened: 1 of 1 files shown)


patrickzheng200 (dismissed):
LGTM

revocation/ocsp/ocsp.go
```diff
@@ -167,7 +167,7 @@ func extensionsToMap(extensions []pkix.Extension) map[string][]byte {
 func executeOCSPCheck(cert, issuer *x509.Certificate, server string, opts Options) (*ocsp.Response, error) {
 	// TODO: Look into other alternatives for specifying the Hash
 	// https://github.com/notaryproject/notation-core-go/issues/139
-	ocspRequest, err := ocsp.CreateRequest(cert, issuer, &ocsp.RequestOptions{Hash: crypto.SHA256})
+	ocspRequest, err := ocsp.CreateRequest(cert, issuer, &ocsp.RequestOptions{Hash: crypto.SHA1})
```
shizhMSFT, line 170: @kody-kimberl Could you put what you have written in the PR description here as a comment block for justification for security auditing?

shizhMSFT (changes requested):

revocation/ocsp/ocsp.go
```diff
@@ -177,7 +177,7 @@ func executeOCSPCheck(cert, issuer *x509.Certificate, server string, opts Option
 		reader := bytes.NewReader(ocspRequest)
 		resp, err = opts.HTTPClient.Post(server, "application/ocsp-request", reader)
 	} else {
-		encodedReq := base64.URLEncoding.EncodeToString(ocspRequest)
+		encodedReq := url.QueryEscape(base64.StdEncoding.EncodeToString(ocspRequest))
```
shizhMSFT, line 180: Do you want to also change line 176?

shizhMSFT: You may also take a look at [url.PathEscape](https://pkg.go.dev/net/url#PathEscape).

revocation/ocsp/ocsp.go
```diff
@@ -167,17 +167,25 @@ func extensionsToMap(extensions []pkix.Extension) map[string][]byte {
...
 	}
 
 	var resp *http.Response
-	if base64.URLEncoding.EncodedLen(len(ocspRequest)) >= 255 {
+	if base64.StdEncoding.EncodedLen(len(ocspRequest)) >= 255 {
 		reader := bytes.NewReader(ocspRequest)
 		resp, err = opts.HTTPClient.Post(server, "application/ocsp-request", reader)
 	} else {
-		encodedReq := base64.URLEncoding.EncodeToString(ocspRequest)
+		encodedReq := url.QueryEscape(base64.StdEncoding.EncodeToString(ocspRequest))
```
shizhMSFT, line 188: `encodedReq` is passed as a path instead of query. I think it would be `PathEscape` instead of `QueryEscape`.

Reference: RFC 2560:

> GET {url}/{url-encoding of base-64 encoding of the DER encoding of
   the OCSPRequest}

revocation/ocsp/ocsp.go
```diff
@@ -167,17 +167,25 @@ func extensionsToMap(extensions []pkix.Extension) map[string][]byte {
...
+	// hashing algorithm SHA1, which has been confirmed to be supported by all
+	// that were tested.
+	ocspRequest, err := ocsp.CreateRequest(cert, issuer, &ocsp.RequestOptions{Hash: crypto.SHA1})
 	if err != nil {
 		return nil, GenericError{Err: err}
 	}
 
 	var resp *http.Response
-	if base64.URLEncoding.EncodedLen(len(ocspRequest)) >= 255 {
+	if base64.StdEncoding.EncodedLen(len(ocspRequest)) >= 255 {
```
shizhMSFT, line 184: The RFC2560 states

> To enable HTTP caching, small requests (that
   after encoding are less than 255 bytes), MAY be submitted using GET.

But I am not sure which encoding it refers to. If it refers to `url-encoding of base-64 encoding of the DER encoding of
the OCSPRequest`, we need an efficient way to calculate the encoded length at least with memory complexity of `O(1)`.

kody-kimberl: My guess is that it is before url-encoding. However, url-encoding would be longer, so I'd say let's use that to be safe. One way that would have memory O(1) but time O(n) would be counting the number of characters that would be escaped and adding 2x that number to the length (since each replacement is with a %XX)

kody-kimberl: Actually, this would require us to store the encoded URL. I think it is best to just check the base64 encoded length first, then only create the path escaped one if that is okay and confirm the length of that. (This would mean it couldn't ever exceed a max size of 255*3 because it would need to be base64 encoded to < 255 and each replacement would be at most 3 characters replacing 1, so technically O(1) since there is an upper bound)

revocation/ocsp/ocsp.go
kody-kimberl: After testing, I can confirm that PathEscape will not work. My guess is that it is a result of URL encoding the "=" characters at the end of base64. We need to use QueryEscape. I have also confirmed that not escaping will also work, so there is not a risk related to lacking URL encoding characters.

kody-kimberl: Resolving since we cannot change this function

shizhMSFT (dismissed):
LGTM

revocation/ocsp/ocsp.go
```diff
@@ -167,17 +167,30 @@ func extensionsToMap(extensions []pkix.Extension) map[string][]byte {
...
+		encodedReq = url.QueryEscape(base64.StdEncoding.EncodeToString(ocspRequest))
+		postRequired = len(encodedReq) >= 255
+	}
+	if postRequired {
 		reader := bytes.NewReader(ocspRequest)
 		resp, err = opts.HTTPClient.Post(server, "application/ocsp-request", reader)
 	} else {
-		encodedReq := base64.URLEncoding.EncodeToString(ocspRequest)
 		var reqURL string
 		reqURL, err = url.JoinPath(server, encodedReq)
```
patrickzheng200, line 195: At line 195, `reqURL, err = url.JoinPath(server, encodedReq)`, where`encodedReq` is defined at line 185 `var encodedReq string`. Logically, this is working, but it's kind of hard to read through this logic because the variable `postRequired` is dancing around and can be toggled between true and false in between.

revocation/ocsp/ocsp.go
```diff
@@ -167,17 +167,30 @@ func extensionsToMap(extensions []pkix.Extension) map[string][]byte {
...
 
 	var resp *http.Response
-	if base64.URLEncoding.EncodedLen(len(ocspRequest)) >= 255 {
+	postRequired := base64.StdEncoding.EncodedLen(len(ocspRequest)) >= 255
+	var encodedReq string
+	if !postRequired {
+		encodedReq = url.QueryEscape(base64.StdEncoding.EncodeToString(ocspRequest))
+		postRequired = len(encodedReq) >= 255
+	}
+	if postRequired {
```
patrickzheng200, line 190: How about:
```suggestion
	postRequired := base64.StdEncoding.EncodedLen(len(ocspRequest)) >= 255
	var encodedReq string
	if !postRequired {
		encodedReq = url.QueryEscape(base64.StdEncoding.EncodeToString(ocspRequest))
		if len(encodedReq) >= 255 {
		    reader := bytes.NewReader(ocspRequest)
		    resp, err = opts.HTTPClient.Post(server, "application/ocsp-request", reader)
		}
	} else {
	    var reqURL string
	    reqURL, err = url.JoinPath(server, encodedReq)
	    ...
	}
```

kody-kimberl: This logic isn't quite right. We want the post to happen for both situations when it is longer (which would require duplication of that code if we nest the if statements). Also, the get should be used for if it is the smallest. I think I can refactor this to try to make it cleaner

patrickzheng200: I have edited my suggestion above.

revocation/ocsp/ocsp.go
kody-kimberl: Updated this logic in the most recent commit

revocation/ocsp/ocsp.go
kody-kimberl: Updated this logic in the most recent commit

revocation/ocsp/ocsp.go
```diff
@@ -167,23 +167,36 @@ func extensionsToMap(extensions []pkix.Extension) map[string][]byte {
...
-	} else {
-		encodedReq := base64.URLEncoding.EncodeToString(ocspRequest)
-		var reqURL string
-		reqURL, err = url.JoinPath(server, encodedReq)
-		if err != nil {
-			return nil, GenericError{Err: err}
+	postRequired := base64.StdEncoding.EncodedLen(len(ocspRequest)) >= 255
+	var encodedReq string
+	if !postRequired {
+		encodedReq = url.QueryEscape(base64.StdEncoding.EncodeToString(ocspRequest))
```
patrickzheng200, line 187: nit:
```suggestion
	if !postRequired {
		encodedReq := url.QueryEscape(base64.StdEncoding.EncodeToString(ocspRequest))
```

patrickzheng200 (approved):
Thanks @kody-kimberl! The logic looks much cleaner now.

priteshbandi (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #169, opened 2023-10-20
fix: update key usage error message to make it clear

### LINKED ISSUE #164: Key usage error message
### What is not working as expected?

> Error: certificate-chain is invalid, invalid self-signed certificate. Error: certificate with subject "CN=xx.com,O=Notary,L=xx,ST=CA,C=US": key usage must not have the bit positions for ContentCommitment, KeyEncipherment, DataEncipherment, KeyAgreement, CertSign, CRLSign, EncipherOnly, DecipherOnly set

If the certificate has more key usages than DigitalSignature, the error message will display all the other key usages. However, this can be confusing if the user didn't set those key usages.

### What did you expect to happen?

The error message should either state that there should be only one DigitalSignature key usage or mention the specific key usages that the certificate actually has and should not exist, instead of all of them.

### How can we reproduce it?

sign with a certificate that has unnecessary key usages

### Describe your environment

Linux

### What is the version of your notation-core-go Library?

v1.0.0

JeyJeyGao:
Resolves #164 

### DIFF
--- a/x509/cert_validations.go
+++ b/x509/cert_validations.go
@@ -203,7 +203,7 @@ func validateLeafKeyUsage(cert *x509.Certificate) error {
 		invalidKeyUsages = append(invalidKeyUsages, "DecipherOnly")
 	}
 	if len(invalidKeyUsages) > 0 {
-		return fmt.Errorf("certificate with subject %q: key usage must not have the bit positions for %s set", cert.Subject, strings.Join(invalidKeyUsages, ", "))
+		return fmt.Errorf("certificate with subject %q is invalid: key usage must be 'Digital Signature' only, found %s", cert.Subject, strings.Join(invalidKeyUsages, ", "))
 	}
 	return nil
 }
--- a/x509/cert_validations.go
+++ b/x509/cert_validations.go
@@ -20,6 +20,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 )
 
@@ -186,8 +187,34 @@ func validateLeafKeyUsage(cert *x509.Certificate) error {
 	if cert.KeyUsage&x509.KeyUsageDigitalSignature == 0 {
 		return fmt.Errorf("certificate with subject %q: key usage must have the bit positions for digital signature set", cert.Subject)
 	}
-	if cert.KeyUsage&kuLeafCertBlocked != 0 {
-		return fmt.Errorf("certificate with subject %q: key usage must not have the bit positions for %s set", cert.Subject, kuLeafCertBlockedString)
+
+	var invalidKeyUsages []string
+	if cert.KeyUsage&x509.KeyUsageContentCommitment != 0 {
+		invalidKeyUsages = append(invalidKeyUsages, "ContentCommitment")
+	}
+	if cert.KeyUsage&x509.KeyUsageKeyEncipherment != 0 {
+		invalidKeyUsages = append(invalidKeyUsages, "KeyEncipherment")
+	}
+	if cert.KeyUsage&x509.KeyUsageDataEncipherment != 0 {
+		invalidKeyUsages = append(invalidKeyUsages, "DataEncipherment")
+	}
+	if cert.KeyUsage&x509.KeyUsageKeyAgreement != 0 {
+		invalidKeyUsages = append(invalidKeyUsages, "KeyAgreement")
+	}
+	if cert.KeyUsage&x509.KeyUsageCertSign != 0 {
+		invalidKeyUsages = append(invalidKeyUsages, "CertSign")
+	}
+	if cert.KeyUsage&x509.KeyUsageCRLSign != 0 {
+		invalidKeyUsages = append(invalidKeyUsages, "CRLSign")
+	}
+	if cert.KeyUsage&x509.KeyUsageEncipherOnly != 0 {
+		invalidKeyUsages = append(invalidKeyUsages, "EncipherOnly")
+	}
+	if cert.KeyUsage&x509.KeyUsageDecipherOnly != 0 {
+		invalidKeyUsages = append(invalidKeyUsages, "DecipherOnly")
+	}
+	if len(invalidKeyUsages) > 0 {
+		return fmt.Errorf("certificate with subject %q: key usage must not have the bit positions for %s set", cert.Subject, strings.Join(invalidKeyUsages, ", "))
 	}
 	return nil
 }
--- a/x509/cert_validations.go
+++ b/x509/cert_validations.go
@@ -174,7 +174,7 @@ func validateLeafKeyUsage(cert *x509.Certificate) error {
 		return err
 	}
 	if cert.KeyUsage&x509.KeyUsageDigitalSignature == 0 {
-		return fmt.Errorf("certificate with subject %q: key usage must have the bit positions for digital signature set", cert.Subject)
+		return fmt.Errorf("The certificate with subject %q is invalid. The key usage must have the bit positions for \"Digital Signature\"", cert.Subject)
 	}
 
 	var invalidKeyUsages []string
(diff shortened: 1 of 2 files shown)


x509/cert_validations.go
```diff
@@ -186,8 +176,34 @@ func validateLeafKeyUsage(cert *x509.Certificate) error {
...
+		invalidKeyUsages = append(invalidKeyUsages, "CRLSign")
+	}
+	if cert.KeyUsage&x509.KeyUsageEncipherOnly != 0 {
+		invalidKeyUsages = append(invalidKeyUsages, "EncipherOnly")
+	}
+	if cert.KeyUsage&x509.KeyUsageDecipherOnly != 0 {
+		invalidKeyUsages = append(invalidKeyUsages, "DecipherOnly")
+	}
+	if len(invalidKeyUsages) > 0 {
+		return fmt.Errorf("certificate with subject %q: key usage must not have the bit positions for %s set", cert.Subject, strings.Join(invalidKeyUsages, ", "))
```
Two-Hearts, line 206: Can we make this error message more user friendly? The user might not fully understand what does `bit position` mean in key usage. Besides providing the error, we should provide information about what Notation is expecting. Something like:
```suggestion
		return fmt.Errorf("certificate with subject %q is invalid: key usage must be 'Digital Signature' only. Found %s", cert.Subject, strings.Join(invalidKeyUsages, ", "))
```

JeyJeyGao: Updated.

x509/cert_validations.go
```diff
@@ -186,8 +176,34 @@ func validateLeafKeyUsage(cert *x509.Certificate) error {
...
+		invalidKeyUsages = append(invalidKeyUsages, "CRLSign")
+	}
+	if cert.KeyUsage&x509.KeyUsageEncipherOnly != 0 {
+		invalidKeyUsages = append(invalidKeyUsages, "EncipherOnly")
+	}
+	if cert.KeyUsage&x509.KeyUsageDecipherOnly != 0 {
+		invalidKeyUsages = append(invalidKeyUsages, "DecipherOnly")
+	}
+	if len(invalidKeyUsages) > 0 {
+		return fmt.Errorf("certificate with subject %q is invalid: key usage must be 'Digital Signature' only, found %s", cert.Subject, strings.Join(invalidKeyUsages, ", "))
```
yizha1, line 206:
```suggestion
		return fmt.Errorf("The certificate with subject %q is invalid. The key usage must be 'Digital Signature' only, but %s was found instead.", cert.Subject, strings.Join(invalidKeyUsages, ", "))
```

yizha1 (commented):
Thanks @JeyJeyGao 

improve the error message a bit.

x509/cert_validations.go
JeyJeyGao: What about "The certificate with subject %q is invalid. The key usage must be \"Digital Signature\" only, but found %s"

Can we avoid using `was` because it is possible that there more than one items.

JeyJeyGao: Updated.

Two-Hearts (dismissed):
LGTM

x509/cert_validations.go
```diff
@@ -184,10 +174,36 @@ func validateLeafKeyUsage(cert *x509.Certificate) error {
 		return err
 	}
 	if cert.KeyUsage&x509.KeyUsageDigitalSignature == 0 {
-		return fmt.Errorf("certificate with subject %q: key usage must have the bit positions for digital signature set", cert.Subject)
+		return fmt.Errorf("The certificate with subject %q is invalid. The key usage must have the bit positions for \"Digital Signature\"", cert.Subject)
```
shizhMSFT, line 177: nit:
```suggestion
		return fmt.Errorf("The certificate with subject %q is invalid. The key usage must have the bit positions for \"Digital Signature\" set", cert.Subject)
```

x509/cert_validations.go
```diff
@@ -184,10 +174,36 @@ func validateLeafKeyUsage(cert *x509.Certificate) error {
...
+	if cert.KeyUsage&x509.KeyUsageEncipherOnly != 0 {
+		invalidKeyUsages = append(invalidKeyUsages, "EncipherOnly")
+	}
+	if cert.KeyUsage&x509.KeyUsageDecipherOnly != 0 {
+		invalidKeyUsages = append(invalidKeyUsages, "DecipherOnly")
 	}
-	if cert.KeyUsage&kuLeafCertBlocked != 0 {
-		return fmt.Errorf("certificate with subject %q: key usage must not have the bit positions for %s set", cert.Subject, kuLeafCertBlockedString)
+	if len(invalidKeyUsages) > 0 {
+		return fmt.Errorf("The certificate with subject %q is invalid. The key usage must be \"Digital Signature\" only, but found %s", cert.Subject, strings.Join(invalidKeyUsages, ", "))
```
shizhMSFT, line 206: `Digital Signature` is double quoted but strings in `invalidKeyUsages` are not.

x509/cert_validations.go
Two-Hearts: hmm, @shizhMSFT I guess we still need to display the current key usage of the invalid certificate?

Two-Hearts: nvm, this is another error message.

x509/cert_validations.go
JeyJeyGao: Added quote for all key usage

x509/cert_validations.go
JeyJeyGao: Updated.

Two-Hearts (dismissed):
LGTM

x509/cert_validations.go
```diff
@@ -184,10 +174,36 @@ func validateLeafKeyUsage(cert *x509.Certificate) error {
 		return err
 	}
 	if cert.KeyUsage&x509.KeyUsageDigitalSignature == 0 {
-		return fmt.Errorf("certificate with subject %q: key usage must have the bit positions for digital signature set", cert.Subject)
+		return fmt.Errorf("The certificate with subject %q is invalid. The key usage must have the bit positions for \"Digital Signature\" set", cert.Subject)
+	}
+
+	var invalidKeyUsages []string
+	if cert.KeyUsage&x509.KeyUsageContentCommitment != 0 {
+		invalidKeyUsages = append(invalidKeyUsages, "\"ContentCommitment\"")
```
shizhMSFT, line 182: nit:
```suggestion
		invalidKeyUsages = append(invalidKeyUsages, `"ContentCommitment"`)
```

shizhMSFT (approved):
LGTM

Two-Hearts (approved):
LGTM

shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM

x509/cert_validations.go
```diff
@@ -184,10 +174,36 @@ func validateLeafKeyUsage(cert *x509.Certificate) error {
 		return err
 	}
 	if cert.KeyUsage&x509.KeyUsageDigitalSignature == 0 {
-		return fmt.Errorf("certificate with subject %q: key usage must have the bit positions for digital signature set", cert.Subject)
+		return fmt.Errorf("The certificate with subject %q is invalid. The key usage must have the bit positions for \"Digital Signature\" set", cert.Subject)
+	}
+
+	var invalidKeyUsages []string
+	if cert.KeyUsage&x509.KeyUsageContentCommitment != 0 {
+		invalidKeyUsages = append(invalidKeyUsages, `"ContentCommitment"`)
```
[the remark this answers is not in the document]
JeyJeyGao: Yes, but if we don't add the quotes here, we'll need a for loop to add quotes for each key's usage. It's a trade-off.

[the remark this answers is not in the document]
JeyJeyGao: Alternatively, we can add it in the join, but it might make the code a bit harder to read.

PR #185, opened 2024-01-20
chore: Adding job step to conditionally execute e2e tests

priteshbandi:
chore: Adding job step to conditionally execute e2e tests

### DIFF
--- a/.github/workflows/reusable-build.yml
+++ b/.github/workflows/reusable-build.yml
@@ -15,13 +15,19 @@ name: Reusable build
 
 on:
   workflow_call:
+    inputs:
+      run-e2e-tests:
+        description: "determines whether to execute end-to-end tests or not"
+        required: false
+        type: boolean
+        default: false
 
 jobs:
   build:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: ["1.20", "1.21"]
+        go-version: [ "1.20", "1.21" ]
       fail-fast: true
     steps:
       - name: Checkout
@@ -33,5 +39,8 @@ jobs:
           check-latest: true
       - name: Run unit tests
         run: make test
+      - name: Run e2e tests
+        if: inputs.run-e2e-tests
+        run: make e2e
       - name: Upload coverage to codecov.io
         uses: codecov/codecov-action@v3


JeyJeyGao:
@priteshbandi Is this for [notation-plugin-framework-go](https://github.com/notaryproject/notation-plugin-framework-go)?
For notation-core-go, it doesn't need an E2E test. I'm not sure if it is appropriate to place the workflow here.

priteshbandi:
> @priteshbandi Is this for [notation-plugin-framework-go](https://github.com/notaryproject/notation-plugin-framework-go)? For notation-core-go, it doesn't need an E2E test. I'm not sure if it is appropriate to place the workflow here.

Yes its for https://github.com/notaryproject/notation-plugin-framework-go/pull/13. I am thinking of expanding reusable workflow to conditionally support support e2e tests.

JeyJeyGao:
[notation-plugin-framework-go](https://github.com/notaryproject/notation-plugin-framework-go): .github/workflows/build.yml

Can we create a new e2e job and depend on build. For example:
```
jobs:
  build:
    uses: notaryproject/notation-core-go/.github/workflows/reusable-build.yml@main

  e2e:
    needs: build
    runs-on: ubuntu-latest
    steps:
      - name: Checkout code
        uses: actions/checkout@v2
      # Add the steps for your e2e test here.
      - name: Run E2E Tests
        run: # Your command to run E2E tests
```
Then, we don't need to update the reusable-build in this repo. @priteshbandi

priteshbandi:
Yes, that works but why do we dont want to add e2e tests in reusable-builds ?

JeyJeyGao:
@priteshbandi I think adding an E2E workflow here will work, but it's not elegant because it will be hard to understand why we need it if one doesn't know that the E2E is for another repository. Are there another other repos also use reusable-builds and need e2e? Or it just be used in [notation-plugin-framework-go](https://github.com/notaryproject/notation-plugin-framework-go)?

priteshbandi:
> @priteshbandi I think adding an E2E workflow here will work, but it's not elegant because it will be hard to understand why we need it if one doesn't know that the E2E is for another repository. Are there another other repos also use reusable-builds and need e2e? Or it just be used in [notation-plugin-framework-go](https://github.com/notaryproject/notation-plugin-framework-go)?

Yes, I agree its not elegant but the name file ` reusable-build.yml` implies that its been used at multiple places. I think we might use (but not sure) e2e for tsa repo which we are creating.

shizhMSFT:
`tspclient-go` does not require E2E tests. Unit tests are sufficient.

`notation` CLI has its own `build.yml` file (see https://github.com/notaryproject/notation/blob/main/.github/workflows/build.yml)

shizhMSFT:
`reusable-build.yml` is intended to be reused by `notation-core-go` and `notation-go`. If this file does not fit your requirement, this file should not be re-used. A new workflow should be created instead.

PR #199, opened 2024-04-07
fix(ci): update codecov token

JeyJeyGao:
The same issue as: https://github.com/notaryproject/notation/issues/921

### DIFF
--- a/.github/workflows/reusable-build.yml
+++ b/.github/workflows/reusable-build.yml
@@ -35,3 +35,5 @@ jobs:
         run: make test
       - name: Upload coverage to codecov.io
         uses: codecov/codecov-action@v4
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}


.github/workflows/reusable-build.yml
```diff
@@ -34,4 +34,6 @@ jobs:
       - name: Run unit tests
         run: make test
       - name: Upload coverage to codecov.io
-        uses: codecov/codecov-action@v4
\ No newline at end of file
+        uses: codecov/codecov-action@v4
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
Two-Hearts, line 39: what's the purpose of this PR? Is there a related issue regarding this?

JeyJeyGao: Added issue link in the PR discription.

Two-Hearts: hmm, according to the doc: https://github.com/codecov/codecov-action?tab=readme-ov-file#breaking-changes:
there's a breaking change on the action, but they also stated that `PRs made from forks to the upstream public repos will support tokenless`. That means we don't need the token?

priteshbandi (approved):
LGTM

.github/workflows/reusable-build.yml
shizhMSFT: Tokenless is only supported in PRs not on the `main` branch

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

Two-Hearts (approved):
LGTM

PR #202, opened 2024-04-10
fix: leaf certificate validation

### LINKED ISSUE #201: bug: leaf certificate key usage should not forbid `ContentCommitment`
### What is not working as expected?

According to Notary Project spec: https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#leaf-certificates,
```
The keyUsage extension MUST be present and MUST be marked critical. Bit positions for digitalSignature MUST be set. The Bit positions for keyEncipherment, dataEncipherment, keyAgreement, keyCertSign, encipherOnly, decipherOnly and cRLSign MUST NOT be set.
```
Key usage `ContentCommitment` (formely known as `nonRepudiation`) is not in the above forbidden list. However, notation-core-go implementation added it in the check.
(From [RFC-5280](https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.3), I'm not seeing any good reason of forbidding `ContentCommitment` while `DigitalSignature` bit is asserted either.)

### What did you expect to happen?

We should remove the `ContentCommitment` check during leaf certificate validation.

### How can we reproduce it?

Use a leaf certificate with `ContentCommitment` and `DigitalSignature` bits asserted.

I met this issue while timestamping my signatures with some free TSAs. This issue blocks those TSAs as they have both `ContentCommitment` and `DigitalSignature` bits asserted in their timestamp signing certificates.

### Describe your environment

all OS, golang v1.22

### What is the version of your notation-core-go Library?

v1.0.2

Two-Hearts:
Resolves #201

### DIFF
--- a/x509/cert_validations.go
+++ b/x509/cert_validations.go
@@ -186,9 +186,6 @@ func validateLeafKeyUsage(cert *x509.Certificate) error {
 	}
 
 	var invalidKeyUsages []string
-	if cert.KeyUsage&x509.KeyUsageContentCommitment != 0 {
-		invalidKeyUsages = append(invalidKeyUsages, `"ContentCommitment"`)
-	}
 	if cert.KeyUsage&x509.KeyUsageKeyEncipherment != 0 {
 		invalidKeyUsages = append(invalidKeyUsages, `"KeyEncipherment"`)
 	}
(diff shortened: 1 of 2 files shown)


shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM

PR #203, opened 2024-04-18
feat: CRL

JeyJeyGao:


revocation/crl/store.go
```diff
@@ -0,0 +1,200 @@
...
+const (
+	// BaseCRL is the file name of the base CRL
+	BaseCRL = "base.crl"
+
+	// Metadata is the file name of the metadata
+	Metadata = "metadata.json"
+)
+
+// Store is an interface to store CRL
+type Store interface {
```
junczhu, line 26: Just for curious,  since we already have `TrustStore` and `CredentialStore`, shall we rename it to `CRLStore`

PR #211, opened 2024-07-15
fix: fix `signerInfo.authenticSigningTime` according to spec

### LINKED ISSUE #38: Support AuthenticSigningTime in SignerInfo
Currently, [`SignerInfo`](https://github.com/notaryproject/notation-core-go/blob/main/signer/signer.go#L32) does not support differentiating SigningTime (incase of `notary.x509`) from AuthenticSigningTime (incase of `notary.signingAuthority.x509`)

Two-Hearts:
This PR fixes `signerInfo.AuthenticSigningTime` to return the correct results. 

According to spec: https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#signing-time--authentic-signing-time: 
`authentic signing time` is a valid concept only under signing scheme [`notary.x509.signingAuthority`](https://github.com/notaryproject/specifications/blob/main/specs/signing-scheme.md/#notaryx509signingauthority). Therefore, for other signing schemes, this function should return an error.

Resolves #38

### DIFF
--- a/signature/types.go
+++ b/signature/types.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"crypto/x509"
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/notaryproject/tspclient-go"
@@ -197,17 +198,17 @@ func (signerInfo *SignerInfo) ExtendedAttribute(key string) (Attribute, error) {
 	return Attribute{}, errors.New("key not in ExtendedAttributes")
 }
 
-// AuthenticSigningTime returns the authentic signing time
+// AuthenticSigningTime returns the authentic signing time under signing scheme
+// notary.x509.signingAuthority.
+// For signing scheme notary.x509, since it only supports authentic timestamp,
+// an error is returned.
+//
+// Reference: https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#signing-time--authentic-signing-time
 func (signerInfo *SignerInfo) AuthenticSigningTime() (time.Time, error) {
-	switch signerInfo.SignedAttributes.SigningScheme {
-	case SigningSchemeX509SigningAuthority:
+	switch signingScheme := signerInfo.SignedAttributes.SigningScheme; {
+	case signingScheme == SigningSchemeX509SigningAuthority:
 		return signerInfo.SignedAttributes.SigningTime, nil
-	case SigningSchemeX509:
-		if len(signerInfo.UnsignedAttributes.TimestampSignature) > 0 {
-			// TODO: Add TSA support for AutheticSigningTime
-			// https://github.com/notaryproject/notation-core-go/issues/38
-			return time.Time{}, errors.New("TSA checking has not been implemented")
-		}
+	default:
+		return time.Time{}, fmt.Errorf("authenticSigningTime not supported under signing scheme %q", signingScheme)
 	}
-	return time.Time{}, errors.New("authenticSigningTime not found")
 }
--- a/signature/types.go
+++ b/signature/types.go
@@ -203,14 +203,15 @@ func (signerInfo *SignerInfo) ExtendedAttribute(key string) (Attribute, error) {
 // For signing scheme notary.x509, since it only supports authentic timestamp,
 // an error is returned.
 //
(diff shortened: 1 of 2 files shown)


signature/types.go
```diff
@@ -197,17 +198,20 @@ func (signerInfo *SignerInfo) ExtendedAttribute(key string) (Attribute, error) {
 	return Attribute{}, errors.New("key not in ExtendedAttributes")
 }
 
-// AuthenticSigningTime returns the authentic signing time
+// AuthenticSigningTime returns the authentic signing time under signing scheme
+// notary.x509.signingAuthority.
+// For signing scheme notary.x509, since it only supports authentic timestamp,
+// an error is returned.
+//
+// Reference: https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#signing-time--authentic-signing-time
```
shizhMSFT, line 206: Instead of referencing the `main` branch, can you reference a permanent link?

signature/types.go
```diff
@@ -197,17 +198,20 @@ func (signerInfo *SignerInfo) ExtendedAttribute(key string) (Attribute, error) {
...
-	switch signerInfo.SignedAttributes.SigningScheme {
-	case SigningSchemeX509SigningAuthority:
-		return signerInfo.SignedAttributes.SigningTime, nil
-	case SigningSchemeX509:
-		if len(signerInfo.UnsignedAttributes.TimestampSignature) > 0 {
-			// TODO: Add TSA support for AutheticSigningTime
-			// https://github.com/notaryproject/notation-core-go/issues/38
-			return time.Time{}, errors.New("TSA checking has not been implemented")
+	switch signingScheme := signerInfo.SignedAttributes.SigningScheme; {
+	case signingScheme == SigningSchemeX509SigningAuthority:
```
shizhMSFT, line 209:
```suggestion
	case SigningSchemeX509SigningAuthority:
```

signature/types.go
```diff
@@ -197,17 +198,20 @@ func (signerInfo *SignerInfo) ExtendedAttribute(key string) (Attribute, error) {
...
+		if signerInfo.SignedAttributes.SigningTime.IsZero() {
+			return time.Time{}, fmt.Errorf("authentic signing time must be present under signing scheme %q", signingScheme)
 		}
+		return signerInfo.SignedAttributes.SigningTime, nil
```
shizhMSFT, line 213: nit: code style: readability: consider
```suggestion
		signingTime := signerInfo.SignedAttributes.SigningTime {
		if signingTime.IsZero() {
			return time.Time{}, fmt.Errorf("authentic signing time must be present under signing scheme %q", signingScheme)
		}
		return signingTime, nil
```

shizhMSFT (approved):
LGTM

signature/types.go
```diff
@@ -197,17 +198,21 @@ func (signerInfo *SignerInfo) ExtendedAttribute(key string) (Attribute, error) {
...
 	case SigningSchemeX509SigningAuthority:
-		return signerInfo.SignedAttributes.SigningTime, nil
-	case SigningSchemeX509:
-		if len(signerInfo.UnsignedAttributes.TimestampSignature) > 0 {
-			// TODO: Add TSA support for AutheticSigningTime
-			// https://github.com/notaryproject/notation-core-go/issues/38
-			return time.Time{}, errors.New("TSA checking has not been implemented")
+		signingTime := signerInfo.SignedAttributes.SigningTime
+		if signingTime.IsZero() {
+			return time.Time{}, fmt.Errorf("authentic signing time must be present under signing scheme %q", signingScheme)
```
priteshbandi, line 212: Is it must under the spec or its optional? AFAIR the previous version of spec said that authentic signing time can be used only under `SigningSchemeX509SigningAuthority` but it was optional

Two-Hearts: @priteshbandi Yes, it's a `REQUIRED` field according to the spec: https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#standard-attributes

One thing to point out is that, we are still using the `SigningTime` field under `SigningSchemeX509SigningAuthority` as authentic signing time. However, based on spec, we should use the `Authentic Signing Time` field instead. (both fields are `REQUIRED`)

priteshbandi (approved):
LGTM

JeyJeyGao (approved):
LGTM

Issue #125, opened 2023-02-28
[revocation] Implement CRL support


priteshbandi:
Spec: https://github.com/notaryproject/notaryproject/blob/main/specs/trust-store-trust-policy.md#crls

Note: For start we can Indirect CRL and can be implemented as step 2.

yizha1:
@priteshbandi  Do you have any plan on implementing CRL checks?

JeyJeyGao:
As we discussed in https://github.com/notaryproject/notation-go/issues/425. We will include the following CRL features for notation **v1.3.0**:
1. **Basic CRL support**  : This includes `base CRL` downloading and validation, with a fallback rule to automatically choose OCSP and fallback to CRL.
2. **CRL cache support**  : Since CRL is a static file with relatively long-term validity, and a well-known CA's CRL is usually greater than 600KiB, a cache mechanism can significantly improve performance.

JeyJeyGao:
CRL and CRL Cache have been merged. We can close this issue now.

PR #214, opened 2024-07-24
feat: CRL

JeyJeyGao:
Feat:
- CRL support, including CRL downloading, validation, and revocation list checks.
- OCSP fallback to CRL when OCSP is unavailable or encounters an unknown issue.

Refactor:
- move OCSP to internal package to export `ocsp.CertCheckStatus` and `ocsp.Supported` function for `revocation` package

The PR implements a subset of RFC 5280:
- It only supports X.509 v2 CRL, as the Golang CRL parser only supports v2.
- It only verifies that the CRL issuer is the certificate's CA with the same key.
- Delta CRL and indirect CRL are not supported.

> NOTE: CRL cache is not included.

Resolves part 1 of #125 


revocation/revocation.go
```diff
@@ -16,26 +16,49 @@
...
+	// and CRL if OCSP is not available. It returns an array of
+	// CertRevocationResults that contain the results and any errors that are
+	// encountered during the process
 	Validate(certChain []*x509.Certificate, signingTime time.Time) ([]*result.CertRevocationResult, error)
 }
 
+// Options specifies values that are needed to check revocation
+type Options struct {
+	// Ctx is a required context used for the revocation check
+	Ctx context.Context
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/revocation.go
```diff
@@ -56,26 +82,141 @@ func NewTimestamp(httpClient *http.Client) (Revocation, error) {
...
+				Result: result.ResultNonRevokable,
+				ServerResults: []*result.ServerResult{{
+					Result: result.ResultNonRevokable,
+					Error:  nil,
+				}},
+			}
+		}
+	}
+
+	// Last is root cert, which will never be revoked by OCSP
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/revocation.go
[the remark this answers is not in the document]
Two-Hearts: I see `context.Background()` is used in `New` and `NewTimestamp` constructors. They should use the context from caller as well. Instead of adding a `ctx` under `revocation` struct, another option is to follow https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/net/http/request.go;l=346 and https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/net/http/request.go;l=362

revocation/revocation.go
```diff
@@ -44,10 +64,12 @@ func New(httpClient *http.Client) (Revocation, error) {
 	if httpClient == nil {
 		return nil, errors.New("invalid input: a non-nil httpClient must be specified")
 	}
-	return &revocation{
-		httpClient:       httpClient,
-		certChainPurpose: ocsp.PurposeCodeSigning,
-	}, nil
+
+	return NewWithOptions(context.Background(), &Options{
```
Two-Hearts, line 68: As mentioned above, `New` should also use the caller's context.

revocation/revocation.go
```diff
@@ -56,26 +78,140 @@ func NewTimestamp(httpClient *http.Client) (Revocation, error) {
...
+		if err := coreX509.ValidateCodeSigningCertChain(certChain, nil); err != nil {
+			return nil, result.InvalidChainError{Err: err}
+		}
+	case ocsp.PurposeTimestamping:
+		if err := coreX509.ValidateTimestampingCertChain(certChain); err != nil {
+			return nil, result.InvalidChainError{Err: err}
+		}
+	default:
+		return nil, result.InvalidChainError{Err: fmt.Errorf("unknown certificate chain purpose %v", r.certChainPurpose)}
+	}
```
Two-Hearts, line 151: This part repeats what we already have under `ocsp.CheckStatus`. So instead of writing it twice, can we refactor it into a helper function that's shared by the two packages?

revocation/revocation.go
```diff
@@ -56,26 +78,140 @@ func NewTimestamp(httpClient *http.Client) (Revocation, error) {
...
+					crlResult := crl.CertCheckStatus(r.ctx, cert, certChain[i+1], crlOpts)
+					crlResult.Error = result.OCSPFallbackError{
+						OCSPErr: ocspResult.Error,
+						CRLErr:  crlResult.Error,
+					}
+					certResults[i] = crlResult
+				} else {
+					certResults[i] = ocspResult
+				}
+			}(i, cert)
```
Two-Hearts, line 187: Ideally, these detailed steps should be done in their corresponding packages.
Under `revocation` package, the `Validate` function should look like this:
```
func (r *revocation) Validate(certChain []*x509.Certificate, signingTime time.Time) ([]*result.CertRevocationResult, error) {
         switch {
         case support OCSP:
              ocsp.CheckStatus(...)
         case support CRL:
              crl.CheckStatus(...)
         }
}
```

revocation/revocation.go
```diff
@@ -56,26 +78,140 @@ func NewTimestamp(httpClient *http.Client) (Revocation, error) {
...
+		default:
+			certResults[i] = &result.CertRevocationResult{
+				Result: result.ResultNonRevokable,
+				ServerResults: []*result.ServerResult{{
+					Result: result.ResultNonRevokable,
+					Error:  nil,
+				}},
+			}
+		}
```
Two-Hearts, line 204: Same as CRL package.

revocation/ocsp/ocsp.go
```diff
@@ -115,15 +115,16 @@ func CheckStatus(opts Options) ([]*result.CertRevocationResult, error) {
 	return certResults, nil
 }
 
-func certCheckStatus(cert, issuer *x509.Certificate, opts Options) *result.CertRevocationResult {
-	ocspURLs := cert.OCSPServer
-	if len(ocspURLs) == 0 {
+// CertCheckStatus checks the revocation status of a certificate using OCSP
+func CertCheckStatus(cert, issuer *x509.Certificate, opts Options) *result.CertRevocationResult {
+	if !SupportOCSP(cert) {
```
Two-Hearts, line 120: Do we really need to export this method? We already have `ocsp.CheckStatus`.

Two-Hearts (commented):
First round of review:
Some comments related to the overall code structure. PTAL.

revocation/revocation.go
Two-Hearts: We will need to deprecated `New`, remove `NewTimestamp`, and add `NewWithOptions` before v1.1.0 stable release.

revocation/revocation.go
```diff
@@ -56,26 +78,140 @@ func NewTimestamp(httpClient *http.Client) (Revocation, error) {
...
+			go func(i int, cert *x509.Certificate) {
+				defer wg.Done()
+				ocspResult := ocsp.CertCheckStatus(cert, certChain[i+1], ocspOpts)
+				if ocspResult != nil && ocspResult.Result == result.ResultUnknown && crl.SupportCRL(cert) {
+					// try CRL check if OCSP result is unknown
+					crlResult := crl.CertCheckStatus(r.ctx, cert, certChain[i+1], crlOpts)
+					crlResult.Error = result.OCSPFallbackError{
+						OCSPErr: ocspResult.Error,
+						CRLErr:  crlResult.Error,
+					}
```
shizhMSFT, line 182: Should you use `errors.Join` instead of `result.OCSPFallbackError`?

revocation/revocation.go
```diff
@@ -56,26 +78,140 @@ func NewTimestamp(httpClient *http.Client) (Revocation, error) {
...
+	var wg sync.WaitGroup
+	for i, cert := range certChain[:len(certChain)-1] {
+		switch {
+		case ocsp.SupportOCSP(cert):
+			// do OCSP check for the certificate
+			wg.Add(1)
 
-	// TODO: add CRL support
-	// https://github.com/notaryproject/notation-core-go/issues/125
+			go func(i int, cert *x509.Certificate) {
```
shizhMSFT, line 173: Instead of passing `i`, could you just pass the issuer cert?

testhelper/certificatetest.go
```diff
@@ -76,15 +76,15 @@ func GetRevokableRSALeafCertificate() RSACertTuple {
 }
 
 // GetRevokableRSAChain returns a chain of certificates that specify a local OCSP server signed using RSA algorithm
-func GetRevokableRSAChain(size int) []RSACertTuple {
+func GetRevokableRSAChain(size int, enabledOCSP, enabledCRL bool) []RSACertTuple {
```
shizhMSFT, line 79: This is also considered as breaking change.

revocation/revocation.go
JeyJeyGao: Updated.

revocation/revocation.go
JeyJeyGao: Waiting for #215

testhelper/certificatetest.go
```diff
@@ -171,31 +171,46 @@ func getRevokableRSACertTuple(cn string, issuer *RSACertTuple) RSACertTuple {
...
 	template.IsCA = true
 	template.KeyUsage = x509.KeyUsageCertSign
-	template.OCSPServer = []string{fmt.Sprintf("http://example.com/chain_ocsp/%d", index)}
+	if enabledOCSP {
+		template.OCSPServer = []string{fmt.Sprintf("http://example.com/chain_ocsp/%d", index)}
+	}
+	if enabledCRL {
+		template.KeyUsage |= x509.KeyUsageCRLSign
+		template.CRLDistributionPoints = []string{fmt.Sprintf("http://example.com/chain_crl/%d", index)}
+
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

testhelper/certificatetest.go
JeyJeyGao: Updated.

Two-Hearts:
A placeholder: this PR depends on https://github.com/notaryproject/notation-core-go/pull/215

Two-Hearts (commented):
A placeholder: this PR depends on https://github.com/notaryproject/notation-core-go/pull/215

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,298 @@
...
+				signingTime.Before(extensions.invalidityDate) {
+				// signing time is before the invalidity date which means the
+				// certificate is not revoked at the time of signing.
+				continue
+			}
+
+			if int(result.CRLReasonCodeCertificateHold) == revocationEntry.ReasonCode ||
+				int(result.CRLReasonCodeRemoveFromCRL) == revocationEntry.ReasonCode {
+				// temporarily revoked
+				tempRevokedEntries = append(tempRevokedEntries, revocationEntry)
```
Two-Hearts, line 192: Do we need to check `revocationEntry.RevocationTime` against zero time?

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,298 @@
...
+	return extensions, nil
+}
+
+func download(ctx context.Context, crlURL string, client *http.Client) (*x509.RevocationList, error) {
+	// validate URL
+	parsedURL, err := url.Parse(crlURL)
+	if err != nil {
+		return nil, fmt.Errorf("invalid CRL URL: %w", err)
+	}
+	if strings.ToLower(parsedURL.Scheme) != "http" {
```
Two-Hearts, line 267:
```suggestion
	if parsedURL.Scheme != "http" {
```
golang net/url package automatically `Parse` the url scheme to lower case.

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,298 @@
...
+		return nil, fmt.Errorf("failed to create CRL request: %w", err)
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("request failed: %w", err)
+	}
+	defer resp.Body.Close()
+
+	// check response
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
```
Two-Hearts, line 283:
```suggestion
	if resp.StatusCode != http.StatusOK {
```

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,298 @@
...
+		return nil, fmt.Errorf("failed to download with status code: %d", resp.StatusCode)
```
Two-Hearts, line 284: Please add the `url` into the error message, so user gets to know which url failed the download.

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	// read with size limit
+	limitedReader := io.LimitReader(resp.Body, maxCRLSize)
+	data, err := io.ReadAll(limitedReader)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read CRL response: %w", err)
+	}
+	if len(data) == maxCRLSize {
+		return nil, fmt.Errorf("CRL size exceeds the limit: %d", maxCRLSize)
```
Two-Hearts, line 294:
```suggestion
		return nil, fmt.Errorf("%s %q: CRL size reached the %d MiB size limit", resp.Request.Method, resp.Request.URL, maxCRLSize/1024/1024)
```

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,298 @@
...
+//
+// If the invalidity date extension is present in the CRL entry and SigningTime
+// is not zero, the certificate is considered revoked if the SigningTime is
+// after the invalidity date. (See RFC 5280, Section 5.3.2)
+func CertCheckStatus(ctx context.Context, cert, issuer *x509.Certificate, opts Options) *result.CertRevocationResult {
+	if !Supported(cert) {
+		return &result.CertRevocationResult{
+			Result: result.ResultNonRevokable,
+			CRLResults: []*result.CRLResult{{
+				Error: fmt.Errorf("CRL is not supported in the certificate %s", cert.Subject.CommonName),
```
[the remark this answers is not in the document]
JeyJeyGao: This is because Golang error messages are often concatenated with other strings, and starting with a lowercase letter makes them flow better in such cases. However, `CRL` is an acronym, so even when concatenated with other strings, the capitalized `CRL` still makes sense.

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,298 @@
...
+			}},
+		}
+	}
+
+	if opts.HTTPClient == nil {
+		opts.HTTPClient = &http.Client{Timeout: 5 * time.Second}
+	}
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/internal/crl/crl.go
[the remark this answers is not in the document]
JeyJeyGao:
```go
	// tempRevokedEntries contains revocation entries with reasons such as
	// CertificateHold or RemoveFromCRL.
	//
	// If the certificate is revoked with CertificateHold, it is temporarily
	// revoked. If the certificate is shown in the CRL with RemoveFromCRL,
	// it is unrevoked.
	var tempRevokedEntries []x509.RevocationListEntry
```
As mentioned here, `CertificateHold` and `RemoveFromCRL` will be stored in `tempRevokedEntries`.

`CertificateHold` means revoked.
`RemoveFromCRL` means canceling the revoke of `CertificateHold`. So, they are a pair of reasons to control the effectiveness of temporarily revoked certificates.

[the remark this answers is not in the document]
JeyJeyGao: Updated the comment:
> // temporarily revoked or unrevoked

JeyJeyGao: In fact, we don't care about the revocation time. As long as the revocation entry exists, it is considered revoked. We do check the `invalidityDate`, which can be used to compare with the `signingTime`.

revocation/internal/crl/crl.go
JeyJeyGao: Nice! Updated.

revocation/internal/crl/crl.go
JeyJeyGao: https://github.com/notaryproject/specifications/blob/main/specs/trust-store-trust-policy.md#crl-download

> If the certificate contains multiple CDP locations then each location download is attempted in sequential order, until a 2xx response is received for any of the location

In our spec, it should be a 2xx response status. Should we follow the spec or not?

revocation/internal/crl/crl.go
JeyJeyGao: Updated.

revocation/internal/crl/crl.go
JeyJeyGao: Updated.

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,298 @@
...
+
+	// The CRLDistributionPoints contains the URIs of all the CRL distribution
+	// points. Since it does not distinguish the reason field, it needs to check
+	// all the URIs to avoid missing any partial CRLs.
+	//
+	// For the majority of the certificates, there is only one CRL distribution
+	// point with one CRL, which will be cached, so checking all the URIs is not
+	// a performance issue.
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/internal/crl/crl.go
Two-Hearts: Well, `resp.StatusCode >= 300` does not follow the spec either?

revocation/result/results.go
```diff
@@ -79,6 +83,71 @@ func NewServerResult(result Result, server string, err error) *ServerResult {
...
+
+const (
+	CRLReasonCodeUnspecified          CRLReasonCode = 0
+	CRLReasonCodeKeyCompromise        CRLReasonCode = 1
+	CRLReasonCodeCACompromise         CRLReasonCode = 2
+	CRLReasonCodeAffiliationChanged   CRLReasonCode = 3
+	CRLReasonCodeSuperseded           CRLReasonCode = 4
+	CRLReasonCodeCessationOfOperation CRLReasonCode = 5
+	CRLReasonCodeCertificateHold      CRLReasonCode = 6
+	// value 7 is not used
```
shizhMSFT, line 97: nit: since the values are explicitly set, it is not necessarily to mention "value 7 is not used" as details can be found in the RFC doc.

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,302 @@
...
+
+	// check if the revocation with CertificateHold or RemoveFromCRL
+	if len(tempRevokedEntries) > 0 {
+		// sort by revocation time (ascending order)
+		sort.Slice(tempRevokedEntries, func(i, j int) bool {
+			return tempRevokedEntries[i].RevocationTime.Before(tempRevokedEntries[j].RevocationTime)
+		})
+
+		// the revocation status depends on the most recent one
+		lastEntry := tempRevokedEntries[len(tempRevokedEntries)-1]
```
shizhMSFT, line 217: You may consider [slices.MaxFunc](https://pkg.go.dev/slices#MaxFunc).

revocation/internal/crl/crl.go
shizhMSFT: Status other than `200 OK` is not useful. For example, if the server returns `204 No Content`, the response has no meanings. We need to fix the spec.

But before the spec is fixed, let's stick with the current spec.

revocation/internal/ocsp/ocsp.go
```diff
@@ -0,0 +1,234 @@
...
+	// Handle pkix-ocsp-no-check and id-ce-invalidityDate extensions if present
+	// in response
+	extensionMap := extensionsToMap(resp.Extensions)
+	if _, foundNoCheck := extensionMap[pkixNoCheckOID]; !foundNoCheck {
+		// This will be ignored until CRL is implemented
+		// If it isn't found, CRL should be used to verify the OCSP response
+		_ = foundNoCheck // needed to bypass linter warnings (Remove after adding CRL)
+		// TODO: add CRL support
+		// https://github.com/notaryproject/notation-core-go/issues/125
+	}
```
shizhMSFT, line 112: Should this part be properly handled by this PR?

revocation/revocation.go
```diff
@@ -101,49 +110,141 @@ func NewWithOptions(opts Options) (Validator, error) {
...
+
+	ocspOpts := ocsp.CertCheckStatusOptions{
+		SigningTime: validateContextOpts.AuthenticSigningTime,
+		HTTPClient:  r.ocspHTTPClient,
+	}
+
+	crlOpts := crl.CertCheckStatusOptions{
+		HTTPClient:  r.crlHTTPClient,
+		SigningTime: validateContextOpts.AuthenticSigningTime,
+	}
```
shizhMSFT, line 175: nit: code style: consistent field name order

revocation/revocation.go
```diff
@@ -101,49 +110,141 @@ func NewWithOptions(opts Options) (Validator, error) {
...
+	wg.Wait()
+
+	// handle panic
+	select {
+	case p := <-panicChan:
+		panic(p)
+	default:
+	}
+	close(panicChan)
```
shizhMSFT, line 247: `panicChan` leaks on panic at line 244.

JeyJeyGao: Updated.

testhelper/certificatetest.go
```diff
@@ -75,16 +75,29 @@ func GetRevokableRSALeafCertificate() RSACertTuple {
 	return revokableRSALeaf
 }
 
+func GetRevokableRSAChainWithRevocations(size int, enabledOCSP, enabledCRL bool) []RSACertTuple {
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/internal/ocsp/ocsp.go
shizhMSFT: Could you open an issue and address it in another PR?

revocation/revocation.go
```diff
@@ -101,49 +110,142 @@ func NewWithOptions(opts Options) (Validator, error) {
...
+		HTTPClient:  r.ocspHTTPClient,
+		SigningTime: validateContextOpts.AuthenticSigningTime,
+	}
+
+	crlOpts := crl.CertCheckStatusOptions{
+		HTTPClient:  r.crlHTTPClient,
+		SigningTime: validateContextOpts.AuthenticSigningTime,
+	}
+
+	panicChan := make(chan any, len(certChain))
```
shizhMSFT, line 177: nit: will it be better to have the following pattern?
```suggestion
	panicChan := make(chan any, len(certChain))
	defer close(panicChan)
```

revocation/internal/ocsp/ocsp.go
JeyJeyGao: Created the issue #220

revocation/revocation.go
JeyJeyGao: Updated.

revocation/revocation.go
```diff
@@ -101,49 +110,142 @@ func NewWithOptions(opts Options) (Validator, error) {
...
+// process.
+//
+// This function tries OCSP and falls back to CRL when:
+// - OCSP is not supported by the certificate
+// - OCSP returns an unknown status
 //
-// TODO: add CRL support
-// https://github.com/notaryproject/notation-core-go/issues/125
+// NOTE: The certificate chain is expected to be in the order of leaf to root.
 func (r *revocation) ValidateContext(ctx context.Context, validateContextOpts ValidateContextOptions) ([]*result.CertRevocationResult, error) {
```
shizhMSFT, line 157: nit: What's the performance of this function? Can it fail fast? If so, please improve it in the next PR.

shizhMSFT (approved):
LGTM

revocation/revocation.go
JeyJeyGao: Created an issue to track the performance optimization #221

revocation/internal/crl/crl.go
```diff
@@ -148,14 +162,30 @@ func validate(crl *x509.RevocationList, issuer *x509.Certificate) error {
...
+		switch {
+		case ext.Id.Equal(oidIssuingDistributionPoint):
+			// IssuingDistributionPoint is a critical extension that identifies
+			// the scope of the CRL. Since we will check all the CRL
+			// distribution points, it is not necessary to check this extension.
+		default:
+			if ext.Critical {
+				return fmt.Errorf("CRL contains unsupported critical extension: %v", ext.Id)
+			}
 		}
```
shizhMSFT, line 174: nit: question: do we need to check other `ext.Id`? If not, I'd prefer
```suggestion
		if ext.Id.Equal(oidIssuingDistributionPoint) {
			// IssuingDistributionPoint is a critical extension that identifies
			// the scope of the CRL. Since we will check all the CRL
			// distribution points, it is not necessary to check this extension.
			continue
		}
		if ext.Critical {
			return fmt.Errorf("CRL contains unsupported critical extension: %v", ext.Id)
		}
```

shizhMSFT (dismissed):
LGTM with suggestions.

revocation/internal/crl/crl.go
JeyJeyGao: Updated.

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,323 @@
...
+		return fmt.Errorf("CRL is expired: %v", crl.NextUpdate)
+	}
+
+	for _, ext := range crl.Extensions {
+		if ext.Id.Equal(oidIssuingDistributionPoint) {
+			// IssuingDistributionPoint is a critical extension that identifies
+			// the scope of the CRL. Since we will check all the CRL
+			// distribution points, it is not necessary to check this extension.
+			continue
+		}
```
Two-Hearts, line 169: According to [RFC 5280 5.2](https://www.rfc-editor.org/rfc/rfc5280#section-5.2):
Conforming CRL issuers are REQUIRED to include the authority key identifier ([Section 5.2.1](https://www.rfc-editor.org/rfc/rfc5280#section-5.2.1)) and the CRL number ([Section 5.2.3](https://www.rfc-editor.org/rfc/rfc5280#section-5.2.3)) extensions in all CRLs issued.

Should we add validations of these two fields here?

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,323 @@
...
+
+	if latestTempRevokedEntry != nil && int(result.CRLReasonCodeRemoveFromCRL) != latestTempRevokedEntry.ReasonCode {
+		// revoked with CertificateHold
+		return &result.CRLResult{
+			Result:         result.ResultRevoked,
+			ReasonCode:     result.CRLReasonCode(latestTempRevokedEntry.ReasonCode),
+			RevocationTime: latestTempRevokedEntry.RevocationTime,
+			URI:            crlURL,
+		}, nil
+	}
```
Two-Hearts, line 247:
```suggestion
	if latestTempRevokedEntry != nil && latestTempRevokedEntry.ReasonCode == int(result.CRLReasonCodeCertificateHold)  {
		// revoked with CertificateHold
		return &result.CRLResult{
			Result:         result.ResultRevoked,
			ReasonCode:     result.CRLReasonCodeCertificateHold,
			RevocationTime: latestTempRevokedEntry.RevocationTime,
			URI:            crlURL,
		}, nil
	}
```

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,323 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// checkRevocation checks if the certificate is revoked or not
+func checkRevocation(cert *x509.Certificate, baseCRL *x509.RevocationList, signingTime time.Time, crlURL string) (*result.CRLResult, error) {
+	if cert == nil {
+		return nil, errors.New("certificate is nil")
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	if baseCRL == nil {
+		return nil, errors.New("CRL is nil")
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/internal/crl/crl.go
JeyJeyGao: Updated.

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,323 @@
...
+			}
+			if len(rest) > 0 {
+				return entryExtensions{}, fmt.Errorf("invalid invalidity date extension: trailing data")
+			}
+
+			extensions.invalidityDate = invalidityDate
+		default:
+			if ext.Critical {
+				// unsupported critical extensions is not allowed. (See RFC 5280, Section 5.2)
+				return entryExtensions{}, fmt.Errorf("CRL entry contains unsupported critical extension: %v", ext.Id)
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/internal/crl/errors.go
```diff
@@ -0,0 +1,22 @@
...
+// limitations under the License.
+
+package crl
+
+import "errors"
+
+var (
+	// ErrDeltaCRLNotChecked is returned when the CRL contains a delta CRL but
+	// the delta CRL are not checked
+	ErrDeltaCRLNotChecked = errors.New("delta CRL are not checked")
```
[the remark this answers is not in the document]
JeyJeyGao: Notation-go may catch this error, then log a warning to indicate that the delta CRL is not checked. Would it be better to name it `ErrDeltaCRLNotChecked`?

revocation/revocation.go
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/revocation.go
```diff
@@ -101,49 +110,142 @@ func NewWithOptions(opts Options) (Validator, error) {
...
+			// do OCSP check for the certificate
+			wg.Add(1)
+
+			go func(i int, cert *x509.Certificate) {
+				defer wg.Done()
+				defer func() {
+					if r := recover(); r != nil {
+						panicChan <- r
+					}
+				}()
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/revocation.go
```diff
@@ -101,49 +110,142 @@ func NewWithOptions(opts Options) (Validator, error) {
...
+
+				ocspResult := ocsp.CertCheckStatus(cert, certChain[i+1], ocspOpts)
+				if ocspResult != nil && ocspResult.Result == result.ResultUnknown && crl.Supported(cert) {
+					// try CRL check if OCSP result is unknown
+					crlResult := crl.CertCheckStatus(ctx, cert, certChain[i+1], crlOpts)
+
+					// insert OCSP result into CRL result
+					crlResult.ServerResults = ocspResult.ServerResults
```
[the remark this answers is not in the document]
JeyJeyGao: I think it had better to be another PR because we will refactor the OCSP package. 
Added an issue #222

revocation/revocation.go
```diff
@@ -101,49 +110,142 @@ func NewWithOptions(opts Options) (Validator, error) {
...
+					Result: result.ResultNonRevokable,
+					Error:  nil,
+				}},
+			}
+		}
+	}
+
+	// Last is root cert, which will never be revoked by OCSP or CRL
+	certResults[len(certChain)-1] = &result.CertRevocationResult{
+		Result: result.ResultNonRevokable,
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/revocation.go
```diff
@@ -101,49 +110,142 @@ func NewWithOptions(opts Options) (Validator, error) {
...
+		}},
+	}
+	wg.Wait()
+
+	// handle panic
+	select {
+	case p := <-panicChan:
+		panic(p)
+	default:
+	}
```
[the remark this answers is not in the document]
JeyJeyGao: No, I think a panic is just that—a panic, and we should rethrow the panic. The issue is more severe than an error. We simply forward the panic from goroutines to the main goroutine to ensure it is not missed.

revocation/revocation.go
```diff
@@ -89,6 +95,9 @@ type Options struct {
 	// OPTIONAL.
 	OCSPHTTPClient *http.Client
 
+	// CRLHTTPClient is a required HTTP client for CRL request
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/result/results.go
```diff
@@ -100,4 +168,7 @@ type CertRevocationResult struct {
 	// Otherwise, every server specified had some error that prevented the
 	// status from being retrieved. These are all contained here for evaluation
 	ServerResults []*ServerResult
```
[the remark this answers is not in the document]
JeyJeyGao: I think it had better to be another PR because we will refactor the OCSP package.
Added an issue https://github.com/notaryproject/notation-core-go/issues/222

revocation/internal/crl/crl.go
Two-Hearts: Yes, the validation is mainly about checking the existence. CRL that missing these fields should be considered invalid and rejected by Notation.

revocation/internal/crl/errors.go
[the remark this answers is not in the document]
Two-Hearts: I guess the confusion comes from the naming itself. Something is `not checked` doesn't necessarily indicate an error. `ErrDeltaCRLNotSupported` sounds like a more proper name for an error?

revocation/internal/crl/crl.go
JeyJeyGao: https://cs.opensource.google/go/go/+/refs/tags/go1.23.0:src/crypto/x509/parser.go;l=1214-1225

The Go built-in `ParseRevocationList` function handles those two extensions, but the library doesn't enforce their existence. I think we should keep it as is without further restrictions, which is the same approach that the Go library takes.

revocation/revocation.go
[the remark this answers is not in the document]
Two-Hearts: However, `crlResult.ServerResults = ocspResult.ServerResults` doesn't look correct to me. This can be easily interpreted as `assigning the server result of ocsp directly to the server result of crl`.

Renaming `crlResult` to `result` makes more sense:
```suggestion
					// insert OCSP result into revocation result
					result.ServerResults = ocspResult.ServerResults
```

revocation/result/results.go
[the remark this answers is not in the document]
Two-Hearts: I'd say we should do it in this iteration, because it's directly related to the `Result`, i.e. it also impacts users of CRL. It should be a quick fix though?

revocation/internal/crl/errors.go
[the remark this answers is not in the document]
Two-Hearts: As we discussed, the revocation check should fail when delta CRL is included because it's not supported in this version. If we simply ignore the delta CRL, we may get a final revocation result with status OK, but in fact the delta CRL contains a newly revoked cert. It should be a rare case, but still brings in security concern.

revocation/revocation.go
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/internal/crl/errors.go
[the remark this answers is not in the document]
JeyJeyGao: Updated to fail the validate when encountered a delta CRL

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,323 @@
...
+//
+// If the invalidity date extension is present in the CRL entry and SigningTime
+// is not zero, the certificate is considered revoked if the SigningTime is
+// after the invalidity date. (See RFC 5280, Section 5.3.2)
+func CertCheckStatus(ctx context.Context, cert, issuer *x509.Certificate, opts CertCheckStatusOptions) *result.CertRevocationResult {
+	if !Supported(cert) {
+		return &result.CertRevocationResult{
+			Result: result.ResultNonRevokable,
+			CRLResults: []*result.CRLResult{{
+				Error: fmt.Errorf("CRL is not supported in the certificate %s", cert.Subject.CommonName),
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/result/results.go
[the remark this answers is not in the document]
JeyJeyGao: Tried to refactor, but it seems redundant to always copy the field from `OCSPResults` to `ServerResults`, which reduces readability. I think we should track this as an issue and rename the field in core-go v2.

revocation/result/results.go
```diff
@@ -90,6 +159,9 @@ type CertRevocationResult struct {
 	// Thus, in the case of more than one ServerResult, this will be ResultUnknown
 	Result Result
 
+	// CRLResults is the result of the CRL check for this certificate
```
Two-Hearts, line 162: Please provide more detailed descriptions of `CRLResults` in the comment, similar to the one of `ServerResults`. For example, when will its length be 0, 1, and/or any other number?

revocation/crl/errors.go
```diff
@@ -0,0 +1,23 @@
...
+
+// Package crl provides the error message for processing crls.
+package crl
+
+import "github.com/notaryproject/notation-core-go/revocation/internal/crl"
+
+var (
+	// ErrDeltaCRLNotSupported is returned when the CRL contains a delta CRL but
+	// the delta CRL is not supported
+	ErrDeltaCRLNotSupported = crl.ErrDeltaCRLNotSupported
```
Two-Hearts, line 22: I think we don't need this exported error anymore? (we can add it back if we need it in the future)

revocation/result/results.go
```diff
@@ -90,6 +159,9 @@ type CertRevocationResult struct {
...
+	CRLResults []*CRLResult
+
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/result/results.go
```diff
@@ -28,8 +32,9 @@ const (
 	// important errors
 	ResultOK
 	// ResultNonRevokable is a Result that indicates that the certificate cannot be
-	// checked for revocation. This may be a result of no OCSP servers being
-	// specified, the cert is a root certificate, or other related situations.
+	// checked for revocation. This may be a result of no OCSP servers or CRL
+	// distribution points being specified, the cert is a root certificate, or
+	// other related situations.
```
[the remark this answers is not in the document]
JeyJeyGao: However, this is not a struct.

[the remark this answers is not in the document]
JeyJeyGao: Also, updated.

Two-Hearts (approved):
LGTM

shizhMSFT (approved):
LGTM

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,306 @@
...
+			// validate signingTime and invalidityDate
+			if !signingTime.IsZero() && !extensions.invalidityDate.IsZero() &&
+				signingTime.Before(extensions.invalidityDate) {
+				// signing time is before the invalidity date which means the
+				// certificate is not revoked at the time of signing.
+				continue
+			}
+
+			switch revocationEntry.ReasonCode {
+			case int(result.CRLReasonCodeCertificateHold), int(result.CRLReasonCodeRemoveFromCRL):
```
priteshbandi, line 204: Shouldn't `removeFromCRL` only appear in delta CRL?

revocation/internal/x509util/validate.go
```diff
@@ -0,0 +1,47 @@
...
+	"fmt"
+
+	"github.com/notaryproject/notation-core-go/revocation/purpose"
+	"github.com/notaryproject/notation-core-go/revocation/result"
+	coreX509 "github.com/notaryproject/notation-core-go/x509"
+)
+
+// ValidateChain checks the certificate chain for a specific purpose, including
+// code signing and timestamping.
+func ValidateChain(certChain []*x509.Certificate, certChainPurpose purpose.Purpose) error {
```
priteshbandi, line 29: nit: please create an issue for this: For revocation we shouldnt care if its time stamping or codesigning cert chain.,

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,306 @@
...
+	// distribution point CRL extension. (See RFC 5280, Section 5.2.5)
+	oidIssuingDistributionPoint = asn1.ObjectIdentifier{2, 5, 29, 28}
+
+	// oidInvalidityDate is the object identifier for the invalidity date
+	// CRL entry extension. (See RFC 5280, Section 5.3.2)
+	oidInvalidityDate = asn1.ObjectIdentifier{2, 5, 29, 24}
+)
+
+// maxCRLSize is the maximum size of CRL in bytes
+const maxCRLSize = 64 * 1024 * 1024 // 64 MiB
```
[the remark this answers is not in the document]
JeyJeyGao: Good question. For large CAs such as Digicert, the CRL is about 800 KiB (http://crl3.digicert.com/DigiCertEVRSACAG2.crl). So, 64 MiB is much larger than that. When we download the CRL, we need to first download it to memory, so we need an appropriate limitation for it. Therefore, 64 MiB is not too large for memory usage and also unreachable for CRL size. Any other better ideas?

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	return &result.CertRevocationResult{
+		Result:     result.ResultOK,
+		CRLResults: crlResults,
+	}
+}
+
+// Supported checks if the certificate supports CRL.
+func Supported(cert *x509.Certificate) bool {
```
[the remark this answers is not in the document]
JeyJeyGao: This is an internal package, so it is only accessible for the revocation package, and external users cannot use it. Yes, we need to export it for the revocation package to check whether the certificate supports CRL.

revocation/result/results.go
```diff
@@ -79,6 +87,70 @@ func NewServerResult(result Result, server string, err error) *ServerResult {
...
+		return "PrivilegeWithdrawn"
+	case CRLReasonCodeAACompromise:
+		return "AACompromise"
+	default:
+		return fmt.Sprintf("invalid reason code with value: %d", r)
+	}
+}
+
+// CRLResult encapsulates the result of a CRL check
+type CRLResult struct {
```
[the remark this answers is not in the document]
JeyJeyGao:
```go

type ServerResult struct {
	// Result of revocation for this server (Unknown if there is an error which
	// prevents the retrieval of a valid status)
	Result Result

	// Server is the URI associated with this result. If no server is associated
	// with the result (e.g. it is a root certificate or no OCSPServers are
	// specified), then this will be an empty string ("")
	Server string

	// Error is set if there is an error associated with the revocation check
	// to this server
	Error error
}

// CRLResult encapsulates the result of a CRL check
type CRLResult struct {
	// Result of revocation for this URI
	Result Result

	// ReasonCode is the reason code for the CRL status
	//
	// The reason code is only set if the certificate was revoked
	ReasonCode CRLReasonCode

	// RevocationTime is the time at which the certificate was revoked
	RevocationTime time.Time

	// URI is the URI to download the CRL
	URI string

	// Error is set if there is an error associated with the revocation check
	Error error
}
```
If we combine them, it will be confusing because `ReasonCode` is only for CRL, and `Server` is only for OCSP. We may need another field, such as `RevocationType`, to distinguish between OCSP and CRL. Also `ServerResult` is not clear name for CRL as CRL is a static file and the result is an analysis from the CRL file. Could you provide a clearer suggestion?

revocation/internal/crl/crl.go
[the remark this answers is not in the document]
priteshbandi: https://chasersystems.com/blog/an-analysis-of-certificate-revocation-list-sizes/

revocation/result/results.go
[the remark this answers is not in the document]
priteshbandi: Both CRL and [OCSP](https://datatracker.ietf.org/doc/html/rfc6960#section-4.2.1) have `ReasonCode`. I think Server is basically an result from one URI (out of many) specified for OCSP checking.

Yes I do see the need of

revocation/internal/crl/crl.go
JeyJeyGao: Good quesiton. According to the RFC 5280: https://www.rfc-editor.org/rfc/rfc5280#section-5.3.1
> The removeFromCRL (8) reasonCode value may only appear in delta CRLs
   and indicates that a certificate is to be removed from a CRL because
   either the certificate expired or was removed from hold.

However, I also reviewed [CA/B forum](https://cabforum.org/uploads/Baseline-Requirements-for-the-Issuance-and-Management-of-Code-Signing.v3.9.pdf)
> 7.2 CRL profile
The serial number of a revoked Certificate MUST remain on the CRL for at least 10 years after the
expiration of the Certificate.

The delta CRL will finally merge with the base CRL when the `NextUpdate` for the base CRL is reached. So, I think the entries for `CertificateHold` and `RemoveFromCRL` should be merged into the base CRL. Also, if the CA doesn't use a Delta CRL, should they update `RemoveFromCRL` directly in the base CRL?

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+func download(ctx context.Context, crlURL string, client *http.Client) (*x509.RevocationList, error) {
+	// validate URL
+	parsedURL, err := url.Parse(crlURL)
+	if err != nil {
+		return nil, fmt.Errorf("invalid CRL URL: %w", err)
+	}
+	if parsedURL.Scheme != "http" {
+		return nil, fmt.Errorf("unsupported scheme: %s. Only supports CRL URL in HTTP protocol", parsedURL.Scheme)
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	// download CRL
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, crlURL, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create CRL request: %w", err)
```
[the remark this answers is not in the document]
JeyJeyGao: Good idea. Updated.

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("request failed: %w", err)
```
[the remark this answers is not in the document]
JeyJeyGao: Good idea. Updated.

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,306 @@
...
+	// check response
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return nil, fmt.Errorf("%s %q: failed to download with status code: %d", resp.Request.Method, resp.Request.URL, resp.StatusCode)
+	}
+
+	// read with size limit
+	limitedReader := io.LimitReader(resp.Body, maxCRLSize)
+	data, err := io.ReadAll(limitedReader)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read CRL response: %w", err)
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/internal/x509util/validate.go
JeyJeyGao: Created #227

revocation/result/results.go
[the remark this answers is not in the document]
JeyJeyGao: But the definition of `ReasonCode` for CRL and OCSP are different, so we cannot share the same field for CRL and OCSP?

revocation/internal/crl/crl.go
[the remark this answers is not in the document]
JeyJeyGao: Updated to 32 MiB

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,306 @@
...
+			if err != nil {
+				return nil, err
+			}
+
+			// validate signingTime and invalidityDate
+			if !signingTime.IsZero() && !extensions.invalidityDate.IsZero() &&
+				signingTime.Before(extensions.invalidityDate) {
+				// signing time is before the invalidity date which means the
+				// certificate is not revoked at the time of signing.
+				continue
```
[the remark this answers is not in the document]
JeyJeyGao: OK, `invalidityDate` extension should not be exsit for `RemoveFromCRL`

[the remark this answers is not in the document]
JeyJeyGao: Updated to use `break`

revocation/result/results.go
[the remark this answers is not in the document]
JeyJeyGao: For example:
```go
type ServerResult struct {
	// Result of revocation for this server (Unknown if there is an error which
	// prevents the retrieval of a valid status)
	Result Result

	// Server is the URI associated with this result. If no server is associated
	// with the result (e.g. it is a root certificate or no OCSPServers are
	// specified), then this will be an empty string ("")
	Server string

	// Error is set if there is an error associated with the revocation check
	// to this server
	Error error

        RevocationMethod Method

	// ReasonCode is the reason code for the CRL status
	//
	// The reason code is only set if the certificate was revoked
	ReasonCode CRLReasonCode
}
```

When `RevocationMethod` == OCSP, the user cannot use `ReasonCode`. Is it what you expected?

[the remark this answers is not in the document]
JeyJeyGao: Is it difficult for the caller to decide which field to use?

revocation/internal/crl/crl.go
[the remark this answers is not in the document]
JeyJeyGao: I found another case:

For example,
```
               Signing Time
                       |
[--------|-------------|-------------|--------------> Time series
         |                           |
   CertificateHold             KeyCompromise
```
Suppose that `KeyCompromise` appears before `CertificateHold` in the revocation list (the revocation list is not ordered). The `invalidityDate` for `KeyCompromise` is after `signingTime`, so we will `break` here based on your suggestion. However, `CertificateHold` occurred before `signingTime`, which means that the key is invalid between the period of `CertificateHold` and `KeyCompromise`. Therefore, we should return that it was revoked. But if we break here, we will return OK. Is this case possible?

revocation/result/results.go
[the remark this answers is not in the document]
priteshbandi: As per [rfc6960#section-4.2.1](https://datatracker.ietf.org/doc/html/rfc6960#section-4.2.1) `revocationReason` is `CRLReason`. So ReasonCode can be used for both OCSP and CRL. Also, the values for is described in [rfc5280#section-5.3.1](https://datatracker.ietf.org/doc/html/rfc5280#section-5.3.1).

revocation/internal/crl/crl.go
priteshbandi: > I think the entries for CertificateHold and RemoveFromCRL should be merged into the base CRL. Also, if the CA doesn't use a Delta CRL, should they update RemoveFromCRL directly in the base CRL?

Can we please confirm this?

revocation/internal/crl/crl.go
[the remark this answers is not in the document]
priteshbandi: I think we need to validate that, a CRL can have multiple entries for same cert? and if it can does  RemoveFromCRL and CertificateHold needs to be in particular order?

revocation/internal/crl/crl.go
JeyJeyGao:
```go
package main

import (
	"context"
	"crypto/x509"
	"fmt"
	"io"
	"net/http"
	"net/url"
)

const maxCRLSize = 32 * 1024 * 1024 // 16 MiB

// download CRL from URL
func download(ctx context.Context, crlURL string, client *http.Client) (*x509.RevocationList, error) {
	// validate URL
	parsedURL, err := url.Parse(crlURL)
	if err != nil {
		return nil, fmt.Errorf("invalid CRL URL: %w", err)
	}
	if parsedURL.Scheme != "http" {
		return nil, fmt.Errorf("unsupported CRL endpoint: %s. Only urls with HTTP scheme is supported", crlURL)
	}

	// download CRL
	req, err := http.NewRequestWithContext(ctx, http.MethodGet, crlURL, nil)
	if err != nil {
		return nil, fmt.Errorf("failed to create CRL request %q: %w", crlURL, err)
	}
	resp, err := client.Do(req)
	if err != nil {
		return nil, fmt.Errorf("request failed for %q: %w", crlURL, err)
	}
	defer resp.Body.Close()

	// check response
	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
		return nil, fmt.Errorf("%s %q: failed to download with status code: %d", resp.Request.Method, resp.Request.URL, resp.StatusCode)
	}

	// read with size limit
	limitedReader := io.LimitReader(resp.Body, maxCRLSize)
	data, err := io.ReadAll(limitedReader)
	if err != nil {
		return nil, fmt.Errorf("failed to read CRL response from %q: %w", resp.Request.URL, err)
	}
	if len(data) == maxCRLSize {
		return nil, fmt.Errorf("%s %q: CRL size reached the %d MiB size limit", resp.Request.Method, resp.Request.URL, maxCRLSize/1024/1024)
	}

	return x509.ParseRevocationList(data)
}

func checkTempRevokedCertificate(crl *x509.RevocationList) {
	for _, entry := range crl.RevokedCertificateEntries {
		if entry.ReasonCode == 6 {
			fmt.Printf("Certificate %s was CertificateHold\n", entry.SerialNumber)
		}

		if entry.ReasonCode == 8 {
			fmt.Printf("Certificate %s was removeFromCRL\n", entry.SerialNumber)
		}
	}
}

func main() {
	testdata := []string{
		"http://crl3.digicert.com/CloudflareIncECCCA-3.crl",
		"http://crl4.digicert.com/CloudflareIncECCCA-3.crl",
		"http://crl.r2m01.amazontrust.com/r2m01.crl",
		"http://crl.r2m02.amazontrust.com/r2m02.crl",
		"http://crl.r2m03.amazontrust.com/r2m03.crl",
		"http://c.pki.goog/wr2/oBFYYahzgVI.crl",
		"http://c.pki.goog/wr2/9UVbN0w5E6Y.crl",
		"http://c.pki.goog/wr2/75r4ZyA3vA0.crl",
		"http://c.pki.goog/wr2/GSyT1N4PBrg.crl",
		"http://c.pki.goog/we1/ygWPENklvpM.crl",
		"http://c.pki.goog/we1/H9bdJBu1Tvg.crl",
		"http://c.pki.goog/we1/T58q3x0jyXI.crl",
		"http://c.pki.goog/we1/FARePxy0M2M.crl",
		"http://c.pki.goog/we1/OhMZjmT1BfY.crl",
		"http://crl3.digicert.com/sha2-ha-server-g6.crl",
		"http://crl4.digicert.com/sha2-ha-server-g6.crl",
		"http://crl3.digicert.com/DigiCertGlobalG2TLSRSASHA2562020CA1-1.crl",
		"http://crl4.digicert.com/DigiCertGlobalG2TLSRSASHA2562020CA1-1.crl",
		"http://crl3.digicert.com/DigiCertTLSHybridECCSHA3842020CA1-1.crl",
		"http://crl4.digicert.com/DigiCertTLSHybridECCSHA3842020CA1-1.crl",
		"http://crl3.digicert.com/DigiCertTLSRSASHA2562020CA1-4.crl",
		"http://crl4.digicert.com/DigiCertTLSRSASHA2562020CA1-4.crl",
		"http://cdp.rapidssl.com/RapidSSLECCCA2018.crl",
		"http://crl3.digicert.com/DigiCertGlobalCAG2.crl",
		"http://crl4.digicert.com/DigiCertGlobalCAG2.crl",
		"http://crl3.digicert.com/DigicertSHA2SecureServerCA-1.crl",
		"http://crl4.digicert.com/DigicertSHA2SecureServerCA-1.crl",
		"http://crl3.digicert.com/DigiCertSecureSiteECCCA-1.crl",
		"http://crl4.digicert.com/DigiCertSecureSiteECCCA-1.crl",
		"http://cdp.geotrust.com/GeoTrustTLSRSACAG1.crl",
		"http://cdp.geotrust.com/GeoTrustRSACA2018.crl",
		"http://crl3.digicert.com/sha2-ev-server-g3.crl",
		"http://crl4.digicert.com/sha2-ev-server-g3.crl",
		"http://cdp.thawte.com/ThawteRSACA2018.crl",
		"http://crl3.digicert.com/DigiCertGlobalG3TLSECCSHA3842020CA1-2.crl",
		"http://crl4.digicert.com/DigiCertGlobalG3TLSECCSHA3842020CA1-2.crl",
		"http://crl3.digicert.com/DigiCertEVRSACAG2.crl",
		"http://crl4.digicert.com/DigiCertEVRSACAG2.crl",
		"http://crl.sectigo.com/SectigoRSAOrganizationValidationSecureServerCA.crl",
		"http://crl.globalsign.com/ca/gsatlasr3dvtlsca2024q1.crl",
		"http://crl.globalsign.com/gsrsaovsslca2018.crl",
		"http://crl.globalsign.com/gsgccr3dvtlsca2020.crl",
		"http://crl.globalsign.com/ca/gsatlasr3dvtlsca2023q4.crl",
		"http://crl.globalsign.com/ca/gsatlasr3dvtlsca2024q2.crl",
		"http://crl.globalsign.com/ca/gsatlasr3dvtlsca2024q3.crl",
		"http://crl.globalsign.com/gseccovsslca2018.crl",
		"http://crl.globalsign.com/alphasslcasha256g4.crl",
		"http://crl.godaddy.com/gdig2s1-8229.crl",
		"http://crl.godaddy.com/gdig2s1-14080.crl",
		"http://www.microsoft.com/pkiops/crl/Microsoft%20Azure%20RSA%20TLS%20Issuing%20CA%2007.crl",
		"http://www.microsoft.com/pkiops/crl/Microsoft%20Azure%20RSA%20TLS%20Issuing%20CA%2008.crl",
		"http://www.microsoft.com/pkiops/crl/Microsoft%20Azure%20ECC%20TLS%20Issuing%20CA%2004.crl",
		"http://crl.apple.com/apevsrsa2g1.crl",
		"http://crl.entrust.net/level1m.crl",
		"http://crl.entrust.net/level1k.crl",
		"http://crl.comodoca.com/COMODOECCOrganizationValidationSecureServerCA.crl",
	}

	for _, url := range testdata {
		crl, err := download(context.Background(), url, http.DefaultClient)
		if err != nil {
			fmt.Println(err)
			return
		}

		fmt.Println("processing CRL:", url)
		checkTempRevokedCertificate(crl)
	}
}
```
I wrote a program to check some CRLs and didn't find `CertificateHold` and `RemoveFromCRL` in the `baseCRL`. Although there is no document forbidding the addition of these two reasons in the base CRL, I cannot find an example to prove they may exist in the base CRL. Let's remove the check for these two reasons and add them back when we implement the delta CRL.

revocation/result/results.go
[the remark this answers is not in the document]
JeyJeyGao: Merged `ServerResult` with `CRLResult`. Also removed the `ReasonCode` and `RevocationTime` as we may don't need them now.

revocation/internal/crl/crl.go
JeyJeyGao: Removed the check for `CertificateHold` and `RemoveFromCRL`.

revocation/method.go
```diff
@@ -0,0 +1,38 @@
...
+
+	// MethodCRL represents CRL as the method used to check the
+	// revocation status of a certificate
+	MethodCRL = crl.RevocationMethodCRL
+
+	// MethodOCSPFallbackCRL represents OCSP check with unknown error
+	// fallback to CRL as the method used to check the revocation status of a
+	// certificate
+	MethodOCSPFallbackCRL = 3
+)
```
Two-Hearts, line 38: Can we make it an enum? So the caller could conveniently call `revocation.Method`:
```suggestion
type Method int

const (
	// MethodUnknown is used for root certificates or when the method
	// used to check the revocation status of a certificate is unknown.
	MethodUnknown Method = iota

	// MethodOCSP represents OCSP as the method used to check the
	// revocation status of a certificate
	MethodOCSP = ocsp.RevocationMethodOCSP

	// MethodCRL represents CRL as the method used to check the
	// revocation status of a certificate
	MethodCRL = crl.RevocationMethodCRL

	// MethodOCSPFallbackCRL represents OCSP check with unknown error
	// fallback to CRL as the method used to check the revocation status of a
	// certificate
	MethodOCSPFallbackCRL = 3
)
```

revocation/result/results.go
```diff
@@ -90,14 +98,28 @@ type CertRevocationResult struct {
 	// Thus, in the case of more than one ServerResult, this will be ResultUnknown
 	Result Result
 
-	// An array of results for each server associated with the certificate.
-	// The length will be either 1 or the number of OCSPServers for the cert.
+	// ServerResults is an array of results for each server associated with the
+	// certificate.
```
Two-Hearts, line 102: Comments for `Result Result` at line 94 to 98 is not accurate anymore since the length of `ServerResult` can be greater than one due to concatenation of OCSP and CRL results.

revocation/result/results.go
```diff
@@ -90,14 +98,28 @@ type CertRevocationResult struct {
...
+	// there is an error, the length will be 1.
+	//
+	// When RevocationMethod is MethodOCSPFallbackCRL, the length
+	// will be the sum of the previous two cases. The CRL result will be
+	// appended after the OCSP results.
 	ServerResults []*ServerResult
+
+	// RevocationMethod is the method used to check the revocation status of the
+	// certificate, including Unknonw(0), MethodOCSP(1), MethodCRL(2) and
+	// OCSPFallbackCRL(3)
```
Two-Hearts, line 123: same here
```suggestion
	// RevocationMethod is the method used to check the revocation status of the
	// certificate, including Unknown(0), MethodOCSP(1), MethodCRL(2) and
	// OCSPFallbackCRL(3)
```

revocation/result/results.go
```diff
@@ -67,6 +71,10 @@ type ServerResult struct {
 	// Error is set if there is an error associated with the revocation check
 	// to this server
 	Error error
+
+	// RevocationMethod is the method used to check the revocation status of the
+	// certificate, including Unknonw(0), MethodOCSP(1), MethodCRL(2)
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/result/results.go
JeyJeyGao: Removed.

revocation/result/results.go
JeyJeyGao: Updated.

revocation/revocation.go
```diff
@@ -101,49 +112,144 @@ func NewWithOptions(opts Options) (Validator, error) {
...
+			}
+		}
+	}
+
+	// Last is root cert, which will never be revoked by OCSP or CRL
+	certResults[len(certChain)-1] = &result.CertRevocationResult{
+		Result: result.ResultNonRevokable,
+		ServerResults: []*result.ServerResult{{
+			Result: result.ResultNonRevokable,
+		}},
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/method.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"github.com/notaryproject/notation-core-go/revocation/internal/crl"
+	"github.com/notaryproject/notation-core-go/revocation/internal/ocsp"
+)
+
+const (
+	// MethodUnknown is used for root certificates or when the method
+	// used to check the revocation status of a certificate is unknown.
+	MethodUnknown int = 0
```
shizhMSFT, line 24: should you make a new `Method` type?
```go
type Method int
```

shizhMSFT (dismissed):
LGTM with suggestions

revocation/internal/crl/crl.go
```diff
@@ -0,0 +1,291 @@
...
+
+	// The CRLDistributionPoints contains the URIs of all the CRL distribution
+	// points. Since it does not distinguish the reason field, it needs to check
+	// all the URIs to avoid missing any partial CRLs.
+	//
+	// For the majority of the certificates, there is only one CRL distribution
+	// point with one CRL URI, which will be cached, so checking all the URIs is
+	// not a performance issue.
+	var (
+		results []*result.ServerResult
```
Two-Hearts, line 93: Let's rename this to `serverResults` and
```suggestion
		serverResults := make([]*result.ServerResult, len(cert.CRLDistributionPoints))
```

revocation/method.go
JeyJeyGao: However, in the `result` package, we cannot call `revocation.Method` type, or we will cause a cycle dependency.

revocation/method.go
```diff
@@ -0,0 +1,38 @@
...
+	MethodOCSP = ocsp.RevocationMethodOCSP
+
+	// MethodCRL represents CRL as the method used to check the
+	// revocation status of a certificate
+	MethodCRL = crl.RevocationMethodCRL
+
+	// MethodOCSPFallbackCRL represents OCSP check with unknown error
+	// fallback to CRL as the method used to check the revocation status of a
+	// certificate
+	MethodOCSPFallbackCRL = 3
```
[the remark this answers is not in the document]
JeyJeyGao: Updated.

revocation/internal/crl/crl.go
JeyJeyGao: Updated.

Two-Hearts (dismissed):
LGTM

priteshbandi (dismissed):
LGTM

revocation/method.go
JeyJeyGao: Added an internal `revocation` package to define the `Method` type.

revocation/method.go
```diff
@@ -14,25 +14,28 @@
 package revocation
 
 import (
-	"github.com/notaryproject/notation-core-go/revocation/internal/crl"
-	"github.com/notaryproject/notation-core-go/revocation/internal/ocsp"
+	internalrevocation "github.com/notaryproject/notation-core-go/revocation/internal/revocation"
```
shizhMSFT, line 17: In `revocation`, you can call `revocation`. There is no naming conflict.
```suggestion
	"github.com/notaryproject/notation-core-go/revocation/internal/revocation"
```

revocation/result/results.go
```diff
@@ -74,7 +78,7 @@ type ServerResult struct {
 
 	// RevocationMethod is the method used to check the revocation status of the
 	// certificate, including Unknown(0), MethodOCSP(1), MethodCRL(2)
-	RevocationMethod int
+	RevocationMethod revocation.Method
```
shizhMSFT, line 81: Here you should reference the exported type in an exported type/method.
```suggestion
	RevocationMethod Method
```

revocation/result/results.go
```diff
@@ -116,6 +120,6 @@ type CertRevocationResult struct {
 
 	// RevocationMethod is the method used to check the revocation status of the
 	// certificate, including Unknown(0), MethodOCSP(1), MethodCRL(2) and
-	// OCSPFallbackCRL(3)
-	RevocationMethod int
+	// MethodOCSPFallbackCRL(3)
```
shizhMSFT, line 123: Numbers should be removed as it is an enum. BTW, should it be `MethodUnknown`?

revocation/result/results.go
```diff
@@ -116,6 +120,6 @@ type CertRevocationResult struct {
...
+	RevocationMethod revocation.Method
```
shizhMSFT, line 124:
```suggestion
	RevocationMethod Method
```

revocation/internal/crl/crl.go
```diff
@@ -90,11 +88,11 @@ func CertCheckStatus(ctx context.Context, cert, issuer *x509.Certificate, opts C
...
 	var (
-		results []*result.ServerResult
-		lastErr error
-		crlURL  string
+		serverResults = make([]*result.ServerResult, 0, len(cert.CRLDistributionPoints))
+		lastErr       error
+		crlURL        string
 	)
-	for _, crlURL = range cert.CRLDistributionPoints {
+	for _, crlURL := range cert.CRLDistributionPoints {
```
shizhMSFT, line 95: Is this change of `:=` intended?

revocation/method.go
JeyJeyGao: image

I tried but it doesn't work.

revocation/result/results.go
JeyJeyGao: However, we cannot call `result.go` is in `result` package and cannot call `Method`.

revocation/result/results.go
JeyJeyGao: Updated.

revocation/result/results.go
JeyJeyGao: However, we cannot call `result.go` is in `result` package and cannot call `Method`.

revocation/internal/crl/crl.go
JeyJeyGao: Nice catch!

revocation/method.go
JeyJeyGao:
```go
// revocation is an internal struct used for revocation checking
type revocation struct {
	ocspHTTPClient   *http.Client
	crlHTTPClient    *http.Client
	certChainPurpose purpose.Purpose
}
```

We have a struct `revocation` in `revocation.go`, which conflicts with this name.

JeyJeyGao: Refactored and moved the `revocation.Method` to `result.RevocationMethod`

revocation/result/results.go
JeyJeyGao: Refactored and moved the `revocation.Method` to `result.RevocationMethod`

revocation/result/results.go
JeyJeyGao: Refactored and moved the `revocation.Method` to `result.RevocationMethod`

shizhMSFT (approved):
LGTM

Two-Hearts (approved):
LGTM

priteshbandi (approved):
LGTM

PR #207, opened 2024-06-20
feat: Timestamp

Two-Hearts:
This PR adds timestamping support with tests.

PR #215, opened 2024-07-29
refactor!: update revocation

Two-Hearts:
This PR updates the revocation package before v1.1.0 stable release. It makes the package more extensible.


revocation/x509/cert.go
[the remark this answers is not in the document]
Two-Hearts: Updated to use `x509.ExtKeyUsage` field.

revocation/revocation.go
```diff
@@ -46,19 +57,25 @@ func New(httpClient *http.Client) (Revocation, error) {
...
 	}, nil
 }
 
-// NewTimestamp contructs a revocation object for timestamping certificate
-// chain
-func NewTimestamp(httpClient *http.Client) (Revocation, error) {
-	if httpClient == nil {
-		return nil, errors.New("invalid input: a non-nil httpClient must be specified")
+// NewWithOptions constructs a revocation object with the specified options
+func NewWithOptions(opts *Options) (Revocation, error) {
```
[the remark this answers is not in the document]
Two-Hearts: Nice catch!

revocation/revocation.go
```diff
@@ -33,10 +35,29 @@ type Revocation interface {
 	Validate(certChain []*x509.Certificate, signingTime time.Time) ([]*result.CertRevocationResult, error)
 }
 
+// ContextRevocation is an interface that specifies methods used for revocation
+// checking with context
+type ContextRevocation interface {
+	// ValidateContext checks the revocation status for a certificate chain
+	// and returns an array of CertRevocationResults that contain the results
+	// and any errors that are encountered during the process
+	ValidateContext(ctx context.Context, certChain []*x509.Certificate, signingTime time.Time) ([]*result.CertRevocationResult, error)
```
JeyJeyGao, line 44: We are creating a new interface, so I want to mention the `revocation mode` feature. Although we will not include it in v1.3.0, will we need it in the future? If we need it, we may consider it now. For `revocation mode`, it cannot be an option for `NewWithOptions` because when we create the revocation client, we don't know which trust policy will be selected, which is based on trust scope. Therefore, the `revocation mode` should be passed in when actually doing the revocation, that is, when calling the `ValidateContext` function. So, should we include `revocation mode` in this function?

JeyJeyGao: `revocation mode` is for selecting the rule of revocation, such as automatic fallback, only OCSP, or only CRL.

JeyJeyGao: Can we provide extensibility for this function. For example
```suggestion
	ValidateContext(ctx context.Context, certChain []*x509.Certificate, signingTime time.Time, opts *ValidateOptions) ([]*result.CertRevocationResult, error)
```

JeyJeyGao: cc: @shizhMSFT

Two-Hearts: @JeyJeyGao I think that's a good idea. We should provide extensibility to the new interface. Let me update the PR.

Two-Hearts: Updated.

revocation/revocation.go
```diff
@@ -46,19 +68,35 @@ func New(httpClient *http.Client) (Revocation, error) {
...
-// chain
-func NewTimestamp(httpClient *http.Client) (Revocation, error) {
-	if httpClient == nil {
-		return nil, errors.New("invalid input: a non-nil httpClient must be specified")
+// Options specifies values that are needed to check revocation
+type Options struct {
+	// OCSPHTTPClient is a required HTTP client for OCSP request
+	OCSPHTTPClient *http.Client
+
+	// CertChainPurpose is the purpose of the certificate chain. Supported
```
JeyJeyGao, line 80: Do we need to mention that this field is required, or we will have a default value for it?

revocation/revocation.go
```diff
@@ -33,10 +35,30 @@ type Revocation interface {
 	Validate(certChain []*x509.Certificate, signingTime time.Time) ([]*result.CertRevocationResult, error)
 }
 
+// ValidateContextOptions provides configuration options for revocation checks
+type ValidateContextOptions struct {
+	// CertChain denotes the certificate chain whose revocation status is
+	// been validated.
+	CertChain []*x509.Certificate
```
JeyJeyGao, line 42: We need to mention that this field is required.

JeyJeyGao (dismissed):
LGTM

revocation/ocsp/ocsp.go
```diff
@@ -78,16 +71,14 @@ func CheckStatus(opts Options) ([]*result.CertRevocationResult, error) {
...
-	case PurposeTimestamping:
+	case x509.ExtKeyUsageTimeStamping:
 		if err := coreX509.ValidateTimestampingCertChain(opts.CertChain); err != nil {
 			return nil, result.InvalidChainError{Err: err}
 		}
 	default:
-		return nil, result.InvalidChainError{Err: fmt.Errorf("unknown certificate chain purpose %v", opts.CertChainPurpose)}
+		if err := coreX509.ValidateCodeSigningCertChain(opts.CertChain, nil); err != nil {
+			return nil, result.InvalidChainError{Err: err}
+		}
```
shizhMSFT, line 81: Setting it to default is dangerous.

Can you modify code and doc saying that setting to x509.ExtKeyUsageAny (which is `0`) implies `CodeSigning`?

revocation/revocation.go
```diff
@@ -33,10 +35,31 @@ type Revocation interface {
...
+}
+
 // revocation is an internal struct used for revocation checking
 type revocation struct {
 	httpClient       *http.Client
-	certChainPurpose ocsp.Purpose
+	certChainPurpose x509.ExtKeyUsage
 }
 
 // New constructs a revocation object for code signing certificate chain
```
shizhMSFT, line 65: Should you mark `New` and `Revocation` as deprecated?

revocation/revocation.go
```diff
@@ -46,19 +69,36 @@ func New(httpClient *http.Client) (Revocation, error) {
...
+	// values are x509.ExtKeyUsageCodeSigning and x509.ExtKeyUsageTimeStamping.
+	// REQUIRED.
+	CertChainPurpose x509.ExtKeyUsage
+}
+
+// NewWithOptions constructs a ContextRevocation with the specified options
+func NewWithOptions(opts Options) (ContextRevocation, error) {
+	if opts.OCSPHTTPClient == nil {
+		return nil, errors.New("invalid input: a non-nil OCSPHTTPClient must be specified")
 	}
```
shizhMSFT, line 91: If http client is not specified, should you use the default http client?

revocation/revocation.go
```diff
@@ -46,19 +69,36 @@ func New(httpClient *http.Client) (Revocation, error) {
...
-		return nil, errors.New("invalid input: a non-nil httpClient must be specified")
+// Options specifies values that are needed to check revocation
+type Options struct {
+	// OCSPHTTPClient is the HTTP client for OCSP request. REQUIRED.
+	OCSPHTTPClient *http.Client
+
+	// CertChainPurpose is the purpose of the certificate chain. Supported
+	// values are x509.ExtKeyUsageCodeSigning and x509.ExtKeyUsageTimeStamping.
+	// REQUIRED.
+	CertChainPurpose x509.ExtKeyUsage
```
shizhMSFT, line 84: If `CertChainPurpose` is not set, it is set to `0`, which is `x509.ExtKeyUsageAny`.

revocation/revocation.go
Two-Hearts: I think we could do it in the next stable release. `notation-go` is still using them. The CRL PR will refactor the `revocation.go` file later on, we can deprecate them in that PR.

revocation/revocation.go
Two-Hearts: Yes. This is in `revocation.go` file, under the `Options` struct, we require the user to provide this field, and the only supported values are `x509.ExtKeyUsageCodeSigning` and `x509.ExtKeyUsageTimeStamping` by design. Under `NewWithOptions`, if caller does not provide their option, it would fail.

Two-Hearts: This is different from the `Options` under the `ocsp` package, where we cannot require the user to provide an option due to breaking change.

revocation/ocsp/ocsp.go
Two-Hearts: Updated.

revocation/revocation.go
Two-Hearts: Updated.

revocation/ocsp/ocsp.go
```diff
@@ -36,24 +36,17 @@ import (
...
 type Options struct {
-	CertChain        []*x509.Certificate
-	CertChainPurpose Purpose // default value is `PurposeCodeSigning`
-	SigningTime      time.Time
-	HTTPClient       *http.Client
+	CertChain []*x509.Certificate
+
+	// CertChainPurpose is the purpose of the certificate chain. Supported
+	// values are x509.ExtKeyUsageCodeSigning and x509.ExtKeyUsageTimeStamping.
+	// When not provided, the chain is taken as a code signing certificate chain.
```
shizhMSFT, line 45: Can you explicitly call out the behavior of `x509.ExtKeyUsageAny`?

revocation/revocation.go
```diff
@@ -16,49 +16,97 @@
...
 
 	"github.com/notaryproject/notation-core-go/revocation/ocsp"
 	"github.com/notaryproject/notation-core-go/revocation/result"
 )
 
-// Revocation is an interface that specifies methods used for revocation checking
+// Revocation is an interface that specifies methods used for revocation checking.
+//
+// Deprecated: Revocation exists for backwards compatibility and should not be used.
+// To perform revocation check, use ContextRevocation.
```
shizhMSFT, line 33: Will `[]` help?
```suggestion
// To perform revocation check, use [ContextRevocation].
```

revocation/revocation.go
```diff
@@ -16,49 +16,97 @@
...
 	httpClient       *http.Client
-	certChainPurpose ocsp.Purpose
+	certChainPurpose x509.ExtKeyUsage
 }
 
-// New constructs a revocation object for code signing certificate chain
+// New constructs a revocation object for code signing certificate chain.
+//
+// Deprecated: New exists for backwards compatibility and should not be used.
+// To create a revocation object, use NewWithOptions.
```
shizhMSFT, line 71: Will `[]` help?
```suggestion
// To create a revocation object, use [NewWithOptions].
```

revocation/revocation.go
```diff
@@ -16,49 +16,97 @@
...
 
-// NewTimestamp contructs a revocation object for timestamping certificate
-// chain
-func NewTimestamp(httpClient *http.Client) (Revocation, error) {
-	if httpClient == nil {
-		return nil, errors.New("invalid input: a non-nil httpClient must be specified")
+// Options specifies values that are needed to check revocation
+type Options struct {
+	// OCSPHTTPClient is the HTTP client for OCSP request. If not provided,
+	// the default http.Client will be used.
```
shizhMSFT, line 85: The term "default http.Client" is vague. You may explicitly say `http.DefaultClient`

revocation/revocation.go
```diff
@@ -16,49 +16,97 @@
...
+	// CertChainPurpose is the purpose of the certificate chain. Supported
+	// values are x509.ExtKeyUsageCodeSigning and x509.ExtKeyUsageTimeStamping.
+	// REQUIRED.
+	CertChainPurpose x509.ExtKeyUsage
+}
+
+// NewWithOptions constructs a ContextRevocation with the specified options
+func NewWithOptions(opts Options) (ContextRevocation, error) {
+	if opts.OCSPHTTPClient == nil {
+		opts.OCSPHTTPClient = &http.Client{}
```
shizhMSFT, line 98:
```suggestion
		opts.OCSPHTTPClient = http.DefaultClient
```

revocation/revocation.go
```diff
@@ -16,49 +16,97 @@
...
+	}
+
+	switch opts.CertChainPurpose {
+	case x509.ExtKeyUsageCodeSigning, x509.ExtKeyUsageTimeStamping:
+	default:
+		return nil, fmt.Errorf("unknown certificate chain purpose %v", opts.CertChainPurpose)
```
shizhMSFT, line 104:
```suggestion
		return nil, fmt.Errorf("unsupported certificate chain purpose %v", opts.CertChainPurpose)
```

shizhMSFT (approved):
LGTM

JeyJeyGao (approved):
LGTM

revocation/revocation.go
```diff
@@ -16,49 +16,97 @@
...
+
+	// AuthenticSigningTime denotes the authentic signing time of the signature.
+	// It is solely used under signing scheme `notary.x509.signingAuthority`.
+	// OPTIONAL.
+	AuthenticSigningTime time.Time
+}
+
+// ContextRevocation is an interface that provides revocation checking with
+// context
+type ContextRevocation interface {
```
priteshbandi, line 55: `ContextRevocation` is an odd name, what does context here means, is it `context.Context`?

revocation/revocation.go
```diff
@@ -16,49 +16,97 @@
...
+// ValidateContextOptions provides configuration options for revocation checks
+type ValidateContextOptions struct {
+	// CertChain denotes the certificate chain whose revocation status is
+	// been validated. REQUIRED.
+	CertChain []*x509.Certificate
+
+	// AuthenticSigningTime denotes the authentic signing time of the signature.
+	// It is solely used under signing scheme `notary.x509.signingAuthority`.
+	// OPTIONAL.
+	AuthenticSigningTime time.Time
```
priteshbandi, line 50: For this package signingTime has nothing to do with `notary.x509.signingAuthority`, its used for revocation check using CRL to determine if signature was generated before certificate was compromised.

revocation/revocation.go
```diff
@@ -16,49 +16,97 @@
...
+	// and returns an array of CertRevocationResults that contain the results
+	// and any errors that are encountered during the process
+	ValidateContext(ctx context.Context, validateContextOpts ValidateContextOptions) ([]*result.CertRevocationResult, error)
+}
+
 // revocation is an internal struct used for revocation checking
 type revocation struct {
 	httpClient       *http.Client
-	certChainPurpose ocsp.Purpose
+	certChainPurpose x509.ExtKeyUsage
```
priteshbandi, line 65: Why do we need purpose, can we determine it from leaf certificate in certificate chain?

revocation/revocation.go
```diff
@@ -16,49 +16,97 @@
...
+	// CertChainPurpose is the purpose of the certificate chain. Supported
+	// values are x509.ExtKeyUsageCodeSigning and x509.ExtKeyUsageTimeStamping.
+	// REQUIRED.
+	CertChainPurpose x509.ExtKeyUsage
+}
+
+// NewWithOptions constructs a ContextRevocation with the specified options
+func NewWithOptions(opts Options) (ContextRevocation, error) {
+	if opts.OCSPHTTPClient == nil {
+		opts.OCSPHTTPClient = http.DefaultClient
```
priteshbandi, line 98: Whats the timeout for `http.DefaultClient`?

revocation/ocsp/ocsp.go
```diff
@@ -36,24 +36,18 @@ import (
...
-	CertChainPurpose Purpose // default value is `PurposeCodeSigning`
-	SigningTime      time.Time
-	HTTPClient       *http.Client
+	CertChain []*x509.Certificate
+
+	// CertChainPurpose is the purpose of the certificate chain. Supported
+	// values are x509.ExtKeyUsageCodeSigning and x509.ExtKeyUsageTimeStamping.
+	// When not provided, the default value x509.ExtKeyUsageAny is also taken as
+	// a code signing certificate chain.
+	CertChainPurpose x509.ExtKeyUsage
```
priteshbandi, line 47: Is the CertChainPurpose field used solely for validating the certificate chain? If so, instead of having a dedicated field for this purpose, could we provide a function that can validate the certificate chain, which the caller will pass? Additionally, could we provide a default implementation of that function, so that callers don't need to worry about the CertChainPurpose field?

revocation/revocation.go
Two-Hearts: Let me rename it to `Validator`. Since the package name is already `revocation`, `revocation.Validator` sounds more reasonable.

revocation/revocation.go
Two-Hearts: This PR is about `revocation` and `ocsp`. 
According to the Notary Project `ocsp` spec: https://github.com/notaryproject/specifications/blob/main/specs/trust-store-trust-policy.md#ocsp section `3.i.a` states that
```
If the invalidity date is present and the authentic signing time is also present then if the invalidity date is before the authentic signing time, the certificate is considered revoked.
```
Therefore, it should be `AuthenticSigningTime` and it's a valid concept solely under `notary.x509.signingAuthority`. For `notary.x509`, we are passing zero value here because it only supports RFC 3161 timestamp.

revocation/revocation.go
Two-Hearts: Updated to `&http.Client{Timeout: 2 * time.Second}`

revocation/ocsp/ocsp.go
Two-Hearts: Tracking in the above comment.

revocation/revocation.go
Two-Hearts: This was introduced in the previous PR (https://github.com/notaryproject/notation-core-go/pull/207) suggested by @shizhMSFT and released in beta.1. So it's not a new field in this PR. We just found that the built-in library is already providing `x509.ExtKeyUsage`, so it's better to use it directly.
Providing a `certChainPurpose` and asking the user to set it `explicitly` adds a guard to prevent the caller accidentally passed in a `code signing cert chain` whereas `timestamping` revocation check is expected, and vice versa.

revocation/revocation.go
priteshbandi: The word `authentic signing time` is misleading its been used to refer to anytimestamping source that user can trust like TSA in case of notary.x509 or  notary.x509.signingAuthority

revocation/revocation.go
priteshbandi: Instead of adding `certChainPurpose` can user pass a cert chain validation function? This way they know what they want.

revocation/revocation.go
Two-Hearts: Okay, let me change it back to `SigningTime` for now. 

Created an issue in the Notary Project spec repo to track this topic: https://github.com/notaryproject/specifications/issues/309

revocation/revocation.go
Two-Hearts: I think the `cert chain validation function` should be controlled by notation-core-go? Because we always want to validate the cert chain during revocation check. If the caller passed in a dummy function, they are allowed to skip the cert chain validation.

For this version, since we already have `certChainPurpose` in beta.1 release, I think we could keep it this way? And we will discuss this further in the v2.0 design of course.

revocation/revocation.go
priteshbandi: to confirm, we are passing signing time when signature is timestamped using TSA?

Two-Hearts: In notation-go, when signing scheme is `notary.x509`, we are passing the zero value, i.e., `time.Time{}` into signing time. In this way, we always say the cert is revoked, `If the certificate is revoked i.e CertStatus is revoked`

shizhMSFT (dismissed):
LGTM

JeyJeyGao (approved):
LGTM

revocation/revocation.go
priteshbandi: With time-stamping we do have autentic signing time.

Two-Hearts: `Timestamping` is a different concept than `Authentic signing time`, we should not mix them. Spec: https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#signing-time--authentic-signing-time

Based on above, we'd better keep the name `SigningTime` in this PR, as it's more general.

priteshbandi: At core they provide similar value. when we know (using timestamping) signature was done before some time, we can use that time and compare it with `InvalidityDate` of CRL to know its revocation status.

Two-Hearts: @priteshbandi Changed back to `AuthenticSigningTime`.

revocation/revocation.go
```diff
@@ -44,9 +44,9 @@ type ValidateContextOptions struct {
 	// been validated. REQUIRED.
 	CertChain []*x509.Certificate
 
-	// SigningTime denotes the signing time of the signature.
+	// AuthenticSigningTime denotes the authentic signing time of the signature.
 	// OPTIONAL.
```
shizhMSFT, line 48: Do you want to add back 
> 	// It is solely used under signing scheme `notary.x509.signingAuthority`.

Two-Hearts: This is comment from Pritesh: https://github.com/notaryproject/notation-core-go/pull/215#discussion_r1704635550

Two-Hearts: Added more comments.

JeyJeyGao (dismissed):
LGTM

revocation/revocation.go
```diff
@@ -16,49 +16,99 @@
...
 	// and any errors that are encountered during the process
 	Validate(certChain []*x509.Certificate, signingTime time.Time) ([]*result.CertRevocationResult, error)
 }
 
+// ValidateContextOptions provides configuration options for revocation checks
+type ValidateContextOptions struct {
+	// CertChain denotes the certificate chain whose revocation status is
+	// been validated. REQUIRED.
+	CertChain []*x509.Certificate
+
```
JeyJeyGao, line 46: Can we add `CertChainPurpose` here instead of `Options`?

JeyJeyGao: Then we don't need 2 validator in notation-go

Two-Hearts: We still have the `Revocatoin` and `New`, although they are deprecated. I think we can make this change in the v2.0 version.

Two-Hearts: Updated.

JeyJeyGao (approved):
LGTM

shizhMSFT (dismissed):
LGTM

priteshbandi (dismissed):

shizhMSFT (approved):
LGTM

JeyJeyGao (approved):
LGTM

priteshbandi (approved):
LGTM

vaninrao10:
LGTM

PR #234, opened 2024-10-08
chore: adding constants defined by notary project spec

Two-Hearts:
Adding constants `MediaTypePayloadV1` and `AnnotationX509ChainThumbprint` to the library as they are defined by notary project spec.

They should be put at `types.go` of notation-core-go along with other types defined by the spec.

Related issue: https://github.com/notaryproject/notation-go/issues/398.

### DIFF
--- a/signature/types.go
+++ b/signature/types.go
@@ -23,14 +23,6 @@ import (
 	"github.com/notaryproject/tspclient-go"
 )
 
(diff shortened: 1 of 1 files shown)


shizhMSFT:
Shouldn't those constants be exported in `notation-go`?

PR #236, opened 2024-10-21
fix: self-signed leaf certificate is not self-issued

### LINKED ISSUE #126: self-signed leaf cert validation issue
We didn’t check the issuer of the self-signed leaf certificate. The self-signed leaf certificate should also be self-issued.
https://github.com/notaryproject/notation-core-go/blob/d336b6a8e953cfe7139234a70a167f883d573bae/x509/cert_validations.go#L46-L58

JeyJeyGao:
Fix:
- added self-issued check for self-signed leaf certificate
- improved test coverage for `x509` package
- fixed test case for `revocation` package

Resolves #126


revocation/revocation_test.go
```diff
@@ -1164,7 +1164,9 @@ func TestCRL(t *testing.T) {
 		}
 
 		revocationClient, err := NewWithOptions(Options{
-			OCSPHTTPClient:   &http.Client{},
+			OCSPHTTPClient: &http.Client{
```
JeyJeyGao, line 1167: Improve test stability.

shizhMSFT (approved):
LGTM

x509/helper.go
```diff
@@ -23,6 +23,19 @@ import (
 	"github.com/notaryproject/notation-core-go/signature"
 )
 
+// validateSelfSignedLeaf validates a self-signed leaf certificate.
+//
+// A self-signed leaf certificate must have the same subject and issuer.
+func validateSelfSignedLeaf(cert *x509.Certificate) error {
```
Two-Hearts, line 29: It seems that we already have an `isSelfSigned` helper function at line 39? Can we make use of it?

JeyJeyGao:
```go
func isSelfSigned(cert *x509.Certificate) (bool, error) {
	return isIssuedBy(cert, cert)
}

func isIssuedBy(subject *x509.Certificate, issuer *x509.Certificate) (bool, error) {
	if err := subject.CheckSignatureFrom(issuer); err != nil {
		return false, err
	}
	return bytes.Equal(issuer.RawSubject, subject.RawIssuer), nil
}
```
We cannot use this function because `CheckSignatureFrom` will enforces `KeyUsageCertSign`, which is required for root cert but not for self-signed leaf cert.

Two-Hearts: I see. Just trying to make concepts clear here. Based on our [spec](https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#other-requirements), `Valid certificate chain MUST contain a root certificate.` 
In this case, for a valid cert chain of length 1, this only cert is still called `root`. (we're just validating it with the `leaf rules`.)

Therefore, for a `self-signed` cert, I think there's only one `valid` scenario in Notation, i.e. `Root cert of a cert chain`.

So a better design might be combining your new function `validateSelfSignedLeaf` into `isSelfSigned`, adding a flag to distinguish between `cert chain length gt 1` and `cert chain length eq 1`.

JeyJeyGao: Updated and consolidated the logic for validating self-signed certificates in the `validateSelfSignedCert` function. Please take a look.

x509/helper.go
```diff
@@ -23,8 +23,34 @@ import (
...
+// Parameters:
+//   - cert: The certificate to validate.
+//   - isSingleCertChain: A boolean value indicating whether the certificate is
+//     the only certificate in the chain.
+//     1. is true: The certificate is both a self-signed leaf and root
+//     certificate.
+//     2. is false: The certificate is a self-signed root CA certificate.
+//
+// Note: A self-signed certificate must have the same subject and issuer.
+func validateSelfSignedCert(cert *x509.Certificate, isSingleCertChain bool) error {
```
Two-Hearts, line 37: On a second thought, do we even need to change method `isSelfSigned`? How about keeping the contents in `helper.go` as is. Because the issue only happens in `cert chain length eq 1` scenario.

x509/codesigning_cert_validations.go
```diff
@@ -34,12 +34,12 @@ func ValidateCodeSigningCertChain(certChain []*x509.Certificate, signingTime *ti
 	// For self-signed signing certificate (not a CA)
 	if len(certChain) == 1 {
 		cert := certChain[0]
+		if err := validateSelfSignedCert(cert, true); err != nil {
+			return err
+		}
```
Two-Hearts, line 39: Can we just add an issuer check here? Something like
```
if !bytes.Equal(cert.RawSubject, cert.RawIssuer) {
    return fmt.Errorf("self-signed certificate with subject %q is not self-issued. The issuer is %q", cert.Subject, cert.Issuer)
}
```
If the logic is used more than one places, we could have a helper function named `isSelfIssued` in `helper.go`.

x509/codesigning_cert_validations.go
```diff
@@ -51,22 +51,18 @@ func ValidateCodeSigningCertChain(certChain []*x509.Certificate, signingTime *ti
...
 		}
 		if i == len(certChain)-1 {
-			selfSigned, selfSignedError := isSelfSigned(cert)
-			if selfSignedError != nil {
-				return fmt.Errorf("root certificate with subject %q is invalid or not self-signed. Certificate chain must end with a valid self-signed root certificate. Error: %v", cert.Subject, selfSignedError)
-			}
-			if !selfSigned {
-				return fmt.Errorf("root certificate with subject %q is not self-signed. Certificate chain must end with a valid self-signed root certificate", cert.Subject)
+			if err := validateSelfSignedCert(cert, false); err != nil {
+				return err
```
Two-Hearts, line 55: After reading through the code again, I think we'd better keep the original `isSelfSigned` logic:
It's used for cert chain with length gt than 1 and the target cert can be leaf/intermediate/root. So it's not assuming the target cert is a root certificate. In fact, `isSelfSigned` does not have any cert chain context, it only checks self-signed.

x509/timestamp_cert_validations.go
```diff
@@ -33,8 +33,8 @@ func ValidateTimestampingCertChain(certChain []*x509.Certificate) error {
 	// For self-signed signing certificate (not a CA)
 	if len(certChain) == 1 {
 		cert := certChain[0]
-		if err := cert.CheckSignature(cert.SignatureAlgorithm, cert.RawTBSCertificate, cert.Signature); err != nil {
-			return fmt.Errorf("invalid self-signed certificate. subject: %q. Error: %w", cert.Subject, err)
+		if err := validateSelfSignedCert(cert, true); err != nil {
+			return err
```
Two-Hearts, line 37: Same here, we could just add the issuer check.
```
if !bytes.Equal(cert.RawSubject, cert.RawIssuer) {
    return fmt.Errorf("self-signed certificate with subject %q is not self-issued. The issuer is %q", cert.Subject, cert.Issuer)
}
```

x509/helper.go
JeyJeyGao: So the original design is fine, and it's just a naming issue with the `validateSelfSignedLeaf` function?

x509/codesigning_cert_validations.go
JeyJeyGao: Updated.

x509/timestamp_cert_validations.go
JeyJeyGao: Updated.

x509/codesigning_cert_validations.go
JeyJeyGao: Updated.

Two-Hearts (approved):
LGTM

x509/codesigning_cert_validations.go
```diff
@@ -34,12 +35,17 @@ func ValidateCodeSigningCertChain(certChain []*x509.Certificate, signingTime *ti
...
 		cert := certChain[0]
-		if signedTimeError := validateSigningTime(cert, signingTime); signedTimeError != nil {
-			return signedTimeError
-		}
+		// check self-signed
 		if err := cert.CheckSignature(cert.SignatureAlgorithm, cert.RawTBSCertificate, cert.Signature); err != nil {
 			return fmt.Errorf("invalid self-signed certificate. subject: %q. Error: %w", cert.Subject, err)
 		}
+		// check self-issued
+		if !bytes.Equal(cert.RawSubject, cert.RawIssuer) {
```
priteshbandi, line 43: Checking issuer == subject is good test but to be sure we need to make sure cert was signed using its private key.

JeyJeyGao: Yes, the truly self-signed check was completed on line 39.

priteshbandi (commented):
LGTM

priteshbandi (approved):

shizhMSFT (approved):
LGTM

x509/codesigning_cert_validations_test.go
```diff
@@ -200,6 +201,35 @@ func TestFailEmptyChain(t *testing.T) {
...
 
+func TestFailNonSelfSignedLeafCert(t *testing.T) {
+	signingTime := time.Now()
+	err := ValidateCodeSigningCertChain([]*x509.Certificate{codeSigningCert}, &signingTime)
+
+	assertErrorEqual("invalid self-signed certificate. subject: \"CN=CodeSigningLeaf\". Error: crypto/rsa: verification error", err, t)
+}
+
+func TestFailSelfIssuedCodeSigningCert(t *testing.T) {
+	chainTuple := testhelper.GetRevokableRSATimestampChain(2)
```
Two-Hearts, line 212: nit:
This is for the code signing cert validation. But a timestamp chain is used in the test.

Two-Hearts: I saw only the leaf cert is used here. Resolving the comment.

PR #238, opened 2024-10-29
fix(test): test cases with `example.com` URL

JeyJeyGao:
Fix:
- added mocked HTTP client for a test case of revocation package
- replaced `*.com` with `*.test` to avoid mentioning a real domain in testing URL

NOTE: `.test` is a reserved domain for testing in RFC 2606

### DIFF
--- a/revocation/crl/fetcher_test.go
+++ b/revocation/crl/fetcher_test.go
@@ -55,7 +55,7 @@ func TestFetch(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to parse base CRL: %v", err)
 	}
-	const exampleURL = "http://example.fake"
+	const exampleURL = "http://localhost.test"
 	const uncachedURL = "http://uncached.com"
 
 	bundle := &Bundle{
@@ -167,7 +167,7 @@ func TestFetch(t *testing.T) {
 			t.Fatalf("failed to parse base CRL: %v", err)
 		}
 		// store the expired CRL
-		const expiredCRLURL = "http://example.fake/expired"
+		const expiredCRLURL = "http://localhost.test/expired"
 		bundle := &Bundle{
 			BaseCRL: expiredCRL,
 		}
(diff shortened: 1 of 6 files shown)


Two-Hearts (commented):
I would suggest replacing any real URLs in test files with fake ones even though we are mocking the servers. This is to avoid future mistaken tests that a developer forgot to mock the server and a real roundtrip got sent out.

JeyJeyGao:
> I would suggest replacing any real URLs in test files with fake ones even though we are mocking the servers. This is to avoid future mistaken tests that a developer forgot to mock the server and a real roundtrip got sent out.

How can we ensure the URL is fake? Maybe the fake one will be registered by someone else later.

JeyJeyGao:
> I would suggest replacing any real URLs in test files with fake ones even though we are mocking the servers. This is to avoid future mistaken tests that a developer forgot to mock the server and a real roundtrip got sent out.

Replaced `example.com` with `example.fake`

Two-Hearts:
> How can we ensure the URL is fake? Maybe the fake one will be registered by someone else later.

Oh, I was thinking we are mocking the server using https://pkg.go.dev/net/http/httptest#Server. Then we could use the `server.URL` directly, turns out we are mocking the http client transport.

shizhMSFT:
@JeyJeyGao RFC 2606 defines some safer domains (see https://datatracker.ietf.org/doc/html/rfc2606#section-2)

JeyJeyGao:
> @JeyJeyGao RFC 2606 defines some safer domains (see https://datatracker.ietf.org/doc/html/rfc2606#section-2)

Updated the URL to be `localhost.test`

Two-Hearts (approved):
LGTM

shizhMSFT:
What about `uncached.com`?

JeyJeyGao:
> What about `uncached.com`?

Nice catch! Updated.

shizhMSFT (approved):
LGTM

Two-Hearts (approved):
LGTM

priteshbandi (approved):
LGTM

PR #243, opened 2024-11-05
fix: timestamping

Two-Hearts:
Due to implementation needs, we must provide a `time.Time` when verifying TSA certificate chain validity after signing. Originally, this time value was taken from the timestamp countersignature envelope itself. 
This PR switches to use the `current time` when verifying TSA certificate chain validity since the tsa cert chain is expected to be valid during signing.


JeyJeyGao (approved):
LGTM

internal/timestamp/timestamp.go
```diff
@@ -43,16 +43,8 @@ func Timestamp(req *signature.SignRequest, opts tspclient.RequestOptions) ([]byt
...
-	if err != nil {
-		return nil, err
-	}
-	timestamp, err := info.Validate(opts.Content)
-	if err != nil {
-		return nil, err
-	}
 	tsaCertChain, err := token.Verify(ctx, x509.VerifyOptions{
-		CurrentTime: timestamp.Value,
+		CurrentTime: req.SigningTime,
```
shizhMSFT, line 47: We should remain the previous implementation due to security consideration we have discussed offline.

Besides, we should have a sanity check on the `req.SigningTime` and the `timestamp.Value` as it only makes sense if the `req.SigningTime` is after the `timestamp.Value`.

shizhMSFT (changes requested):

internal/timestamp/timestamp.go
Two-Hearts: Updated.

shizhMSFT (approved):
LGTM

JeyJeyGao (approved):
LGTM

priteshbandi:
>This PR switches to use the current time when verifying TSA certificate chain validity since the tsa cert chain is expected to be valid during signing.

Is this true for verification ?

Two-Hearts:
> Is this true for verification ?

This PR is for signing only. We will have another PR for verification in notation-go once this PR gets merged.

priteshbandi (approved):
LGTM

PR #246, opened 2024-11-21
fix: add tsa cert chain revocation check after timestamping

### LINKED ISSUE #237: TSA certificate chain revocation check after timestamping
We should consider adding an option for users to enable tsa cert chain revocation check after timestamping.
Currently this is done during verification only, but we can fail fast on the signing side.

Two-Hearts:
This PR adds an option to enable tsa certificate chain revocation check after timestamping during signing.

Resolves #237


internal/timestamp/timestamp.go
```diff
@@ -52,5 +55,59 @@ func Timestamp(req *signature.SignRequest, opts tspclient.RequestOptions) ([]byt
...
+		if certResult.Result == result.ResultOK || certResult.Result == result.ResultNonRevokable {
+			numOKResults++
+		} else {
+			finalResult = certResult.Result
+			problematicCertSubject = cert.Subject.String()
+			if certResult.Result == result.ResultRevoked {
+				revokedFound = true
+				revokedCertSubject = problematicCertSubject
+			}
+		}
```
JeyJeyGao, line 93:
```suggestion
		} else {
			finalResult = certResult.Result
			problematicCertSubject = cert.Subject.String()
			if certResult.Result == result.ResultRevoked {
				revokedFound = true
				revokedCertSubject = problematicCertSubject
			}
			break
		}
```
Can we break here?

Two-Hearts: We actually had the same discussion when we were implementing OCSP, see details here: https://github.com/notaryproject/notation-go/pull/295#discussion_r1169403270.

Resolving based on offline discussion.

JeyJeyGao (approved):
LGTM

x509/helper.go
shizhMSFT, line 1: Is this file changed just for resolving the import cycle? If it is, we need to revisit all the changes.

signature/types.go
```diff
@@ -112,6 +113,11 @@ type SignRequest struct {
 	// TSARootCAs is the set of caller trusted TSA root certificates
 	TSARootCAs *x509.CertPool
 
+	// RevocationTimestampingValidator is used for timestamping certificate
+	// chain revocation check after signing.
+	// When present, only used when timestamping is performed.
+	RevocationTimestampingValidator revocation.Validator
```
shizhMSFT, line 119: Will `TSARevocationValidator` be a better name and consistent?

x509/helper.go
Two-Hearts: It's not only about import cycle, but also following our spec: https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#algorithm-selection

signature/types.go
Two-Hearts: Updated.

x509/helper.go
Two-Hearts: Based on offline discussion, will extract the core logic to make the code cleaner.

Two-Hearts: Updated.

shizhMSFT (dismissed):
LGTM

JeyJeyGao (dismissed):
LGTM

internal/timestamp/timestamp.go
```diff
@@ -52,5 +56,54 @@ func Timestamp(req *signature.SignRequest, opts tspclient.RequestOptions) ([]byt
 	if err := nx509.ValidateTimestampingCertChain(tsaCertChain); err != nil {
 		return nil, err
 	}
+	// certificate chain revocation check after timestamping
+	if req.TSARevocationValidator != nil {
+		certResults, err := req.TSARevocationValidator.ValidateContext(ctx, revocation.ValidateContextOptions{
+			CertChain: tsaCertChain,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("after timestamping: failed to check timestamping certificate chain revocation with error: %w", err)
```
priteshbandi, line 65:
```suggestion
			return nil, fmt.Errorf("failed to validate the revocation status of timestamping certificate(s) with error: %w", err)
```

internal/timestamp/timestamp.go
```diff
@@ -52,5 +56,54 @@ func Timestamp(req *signature.SignRequest, opts tspclient.RequestOptions) ([]byt
...
+		}
+		if err := revocationFinalResult(certResults, tsaCertChain); err != nil {
+			return nil, fmt.Errorf("after timestamping: %w", err)
```
priteshbandi, line 68:
```suggestion
			return nil, err
```

the timestamp certificate revocation check occurs after the timestamping process, making the additional context (`after time stamping`) unnecessary in the error message.

internal/timestamp/timestamp.go
```diff
@@ -52,5 +56,54 @@ func Timestamp(req *signature.SignRequest, opts tspclient.RequestOptions) ([]byt
...
+		}
+	}
 	return resp.TimestampToken.FullBytes, nil
 }
+
+// revocationFinalResult returns an error if any cert in the cert chain has
+// a revocation status other than ResultOK or ResultNonRevokable.
+// When ResultRevoked presents, always return the revoked error.
+func revocationFinalResult(certResults []*result.CertRevocationResult, certChain []*x509.Certificate) error {
```
priteshbandi, line 77: Can we rename `revocationFinalResult` to some thing like `revocationResult` ? `Final` in the name doesn't improves readability.

internal/timestamp/timestamp.go
```diff
@@ -52,5 +56,54 @@ func Timestamp(req *signature.SignRequest, opts tspclient.RequestOptions) ([]byt
...
+			}
+			if !hasUnknownResult { // unknown
+				// not returning because a following cert can be revoked
+				problematicCertSubject = cert.Subject.String()
+				hasUnknownResult = true
+			}
+		}
+	}
+	if numOKResults != len(certResults) {
+		return fmt.Errorf("timestamping certificate with subject %q revocation status is unknown", problematicCertSubject)
```
priteshbandi, line 106:
```suggestion
		return fmt.Errorf("revocation status of timestamping certificate with subject %q is unknown", problematicCertSubject)
```

Two-Hearts: I suggest keeping this one because put `revocation status is unknown` together has a better readability as sometimes the subject in between can be very long.

internal/timestamp/timestamp.go
Two-Hearts: Updated.

internal/timestamp/timestamp.go
Two-Hearts: Updated.

internal/timestamp/timestamp.go
Two-Hearts: Updated.

priteshbandi (approved):
LGTM

JeyJeyGao (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #259, opened 2025-02-18
fix: use iterator instead of looping through multiple slices

JeyJeyGao:
Fix:
- updated CRL revocation list check to use iterator rather than looping through multiple slices

Test:
- added more test cases

Note: the revocation list may contain millions of entries, so the implementation needs to consider memory allocation.

### DIFF
--- a/revocation/internal/crl/crl.go
+++ b/revocation/internal/crl/crl.go
@@ -280,15 +280,15 @@ func checkRevocation(cert *x509.Certificate, b *crl.Bundle, signingTime time.Tim
 	}
 
 	// merge the base and delta CRLs in a single iterator
-	revocationListIter := func(yield func(x509.RevocationListEntry) bool) {
-		for _, revocationEntry := range b.BaseCRL.RevokedCertificateEntries {
-			if !yield(revocationEntry) {
+	revocationListIter := func(yield func(*x509.RevocationListEntry) bool) {
+		for i := range b.BaseCRL.RevokedCertificateEntries {
+			if !yield(&b.BaseCRL.RevokedCertificateEntries[i]) {
 				return
 			}
 		}
 		if b.DeltaCRL != nil {
-			for _, revocationEntry := range b.DeltaCRL.RevokedCertificateEntries {
-				if !yield(revocationEntry) {
+			for i := range b.DeltaCRL.RevokedCertificateEntries {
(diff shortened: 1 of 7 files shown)


shizhMSFT (changes requested):
Can you split this PR into two PRs? One for golang bump up, another for the CRL. Basically, one PR one purpose.

revocation/internal/crl/crl.go
```diff
@@ -279,9 +279,20 @@ func checkRevocation(cert *x509.Certificate, b *crl.Bundle, signingTime time.Tim
 		return nil, errors.New("baseCRL cannot be nil")
 	}
 
-	revocationListBundle := [][]x509.RevocationListEntry{b.BaseCRL.RevokedCertificateEntries}
-	if b.DeltaCRL != nil {
-		revocationListBundle = append(revocationListBundle, b.DeltaCRL.RevokedCertificateEntries)
+	// merge the base and delta CRLs in a single iterator
+	revocationListIter := func(yield func(x509.RevocationListEntry) bool) {
```
shizhMSFT, line 283: Will it be better to do
```suggestion
	revocationListIter := func(yield func(*x509.RevocationListEntry) bool) {
```

shizhMSFT (approved):
LGTM

revocation/internal/crl/crl.go
```diff
@@ -279,9 +279,20 @@ func checkRevocation(cert *x509.Certificate, b *crl.Bundle, signingTime time.Tim
 		return nil, errors.New("baseCRL cannot be nil")
 	}
 
-	revocationListBundle := [][]x509.RevocationListEntry{b.BaseCRL.RevokedCertificateEntries}
-	if b.DeltaCRL != nil {
-		revocationListBundle = append(revocationListBundle, b.DeltaCRL.RevokedCertificateEntries)
```
Two-Hearts, line 284: Why do we need this change?

JeyJeyGao: The requirement is to traverse both the base CRL revocation list and the delta CRL revocation list. Previously, in Go v1.22, we didn't have `iter` support, so we could only merge these two lists into a `revocationListBundle`. However, this approach is awkward and requires an additional loop when traversing these lists.

Two-Hearts (approved):
LGTM

yizha1 (approved):
LGTM

PR #268, opened 2025-04-10
feat: add support of COSE hash envelope

ghost:
This PR targets on the `feature/cose-hash-envelope` branch.

Conformance test is included as well.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,9 +15,13 @@ name: Build
 
 on:
   push:
-    branches: main
+    branches: 
+      - main
+      - cose*
   pull_request:
-    branches: main
+    branches: 
+      - main
+      - cose*
 
 jobs:
   build:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,11 +17,11 @@ on:
   push:
     branches: 
       - main
-      - cose*
+      - feature/*
   pull_request:
     branches: 
       - main
-      - cose*
+      - feature/*
 
 jobs:
   build:
--- a/signature/cose/envelope.go
+++ b/signature/cose/envelope.go
@@ -47,7 +47,6 @@ func init() {
 	if err != nil {
 		panic(err)
 	}
-
 	encOpts := cbor.EncOptions{
 		Time:    cbor.TimeUnix,
 		TimeTag: cbor.EncTagRequired,
@@ -56,7 +55,6 @@ func init() {
 	if err != nil {
 		panic(err)
 	}
-
 	decOpts := cbor.DecOptions{
 		TimeTag: cbor.DecTagRequired,
 	}
@@ -185,7 +183,9 @@ func (signer *localSigner) CertificateChain() []*x509.Certificate {
 }
 
 type envelope struct {
-	base *cose.Sign1Message
+	base               *cose.Sign1Message
+	isCoseHashEnvelope bool   // set to true if base represents a COSE hash envelope
+	raw                []byte // raw COSE hash envelope bytes
 }
 
 // NewEnvelope initializes an empty COSE signature envelope.
@@ -195,11 +195,23 @@ func NewEnvelope() signature.Envelope {
 	}
 }
 
-// ParseEnvelope parses envelopeBytes to a COSE signature envelope.
+// ParseEnvelope parses envelopeBytes to a COSE signature envelope or
+// a COSE hash envelope.
 func ParseEnvelope(envelopeBytes []byte) (signature.Envelope, error) {
 	var msg cose.Sign1Message
 	if err := msg.UnmarshalCBOR(envelopeBytes); err != nil {
 		return nil, &signature.InvalidSignatureError{Msg: err.Error()}
+	}
+	if _, ok := msg.Headers.Protected[cose.HeaderLabelPayloadHashAlgorithm]; ok {
+		// COSE hash envelope
+		return &base.Envelope{
+			Envelope: &envelope{
+				base:               &msg,
+				isCoseHashEnvelope: true,
+				raw:                envelopeBytes,
+			},
+			Raw: envelopeBytes,
+		}, nil
 	}
 	return &base.Envelope{
 		Envelope: &envelope{
@@ -211,6 +223,7 @@ func ParseEnvelope(envelopeBytes []byte) (signature.Envelope, error) {
 
 // Sign implements signature.Envelope interface.
 // On success, this function returns the COSE signature envelope byte slice.
+// When req.CoseHashEnvelope is true, it returns a COSE hash envelope byte slice.
 func (e *envelope) Sign(req *signature.SignRequest) ([]byte, error) {
 	// get built-in signer from go-cose or remote signer based on req.Signer
 	signer, err := getSigner(req.Signer)
@@ -226,10 +239,6 @@ func (e *envelope) Sign(req *signature.SignRequest) ([]byte, error) {
 	if err := generateProtectedHeaders(req, msg.Headers.Protected); err != nil {
 		return nil, &signature.InvalidSignRequestError{Msg: err.Error()}
 	}
-
-	// generate payload of COSE envelope
-	msg.Headers.Protected[cose.HeaderLabelContentType] = req.Payload.ContentType
-	msg.Payload = req.Payload.Content
 
 	// generate unprotected headers of COSE envelope.
 	generateUnprotectedHeaders(req, signer, msg.Headers.Unprotected)
@@ -253,19 +262,15 @@ func (e *envelope) Sign(req *signature.SignRequest) ([]byte, error) {
 		msg.Headers.Unprotected[headerLabelTimestampSignature] = timestampToken
 	}
 
-	// core sign process, generate signature of COSE envelope
-	// blob signing
-	if req.BlobSign {
-		// blob signing returns a COSE hash envelope
-		hashEnvelopePayload := cose.HashEnvelopePayload{
-			HashAlgorithm:       signer.Algorithm(),
-			HashValue:           req.Payload.Content,
-			PreimageContentType: req.Payload.BlobContentMediaType,
-		}
-		return cose.SignHashEnvelope(rand.Reader, signer, msg.Headers, hashEnvelopePayload)
+	// core sign process
+	if req.CoseHashEnvelope { // COSE hash envelope is required
+		return cose.SignHashEnvelope(rand.Reader, signer, msg.Headers, req.CoseHashEnvelopePayload)
 	}
 
-	// oci signing
+	// COSE envelope
+	// generate payload of COSE envelope
+	msg.Headers.Protected[cose.HeaderLabelContentType] = req.Payload.ContentType
+	msg.Payload = req.Payload.Content
 	if err := msg.Sign(rand.Reader, nil, signer); err != nil {
 		return nil, &signature.InvalidSignRequestError{Msg: err.Error()}
 	}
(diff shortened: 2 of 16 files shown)


signature/cose/envelope.go
```diff
@@ -297,27 +321,69 @@ func (e *envelope) Verify() (*signature.EnvelopeContent, error) {
...
 	}
-	err = e.base.Verify(nil, verifier)
-	if err != nil {
-		return nil, &signature.SignatureIntegrityError{Err: err}
+	if e.isCoseHashEnvelope { // verify integrity of COSE hash envelope
+		hashEnv, err := e.base.MarshalCBOR()
+		if err != nil {
+			return nil, &signature.SignatureIntegrityError{Err: err}
+		}
+		if _, err := cose.VerifyHashEnvelope(verifier, hashEnv); err != nil {
```
JeyJeyGao, line 322: I'm not sure about the CBOR marshal and unmarshal performance, but it seems inefficient to marshal and unmarshal the CBOR content. Is it possible to export an API to verify a Sign1Message as a CoseHashEnvelope? Currently, we only support verifying CoseHashEnvelope binary. cc: @shizhMSFT

.github/workflows/build.yml
shizhMSFT, line 1: Can you split the PR for those workflows?

ghost: Done.

signature/cose/envelope.go
ghost: Yeah, but a `cose.Sign1Message` is required at the end of sign. Therefore, the `unmarshal` is needed anyway. So, it seems to be a good idea to introduce a new method for struct `cose.Sign1Message` named `SignHashEnvelope`.

shizhMSFT: We can always tune the performance later. Let's make sure the functionality works first.

signature/types.go
```diff
@@ -118,6 +123,20 @@ type SignRequest struct {
...
+	// It is ignored when signing under the JWS format.
+	CoseHashEnvelope bool
+
+	// CoseHashEnvelopePayload is the payload to be signed and REQUIRED
+	// when [SignRequest.CoseHashEnvelope] is set to true.
+	//
+	// It is ignored when signing under the JWS format or
+	// [SignRequest.CoseHashEnvelope] is set to false.
+	// When ignored, [SignRequest.Payload] is the payload to be signed.
+	CoseHashEnvelopePayload cose.HashEnvelopePayload
```
shizhMSFT: We can move this field into the `Payload`. Also, we can use a pointer to avoid `CoseHashEnvelope bool`.

signature/cose/envelope.go
```diff
@@ -185,7 +183,8 @@ func (signer *localSigner) CertificateChain() []*x509.Certificate {
 }
 
 type envelope struct {
-	base *cose.Sign1Message
+	base               *cose.Sign1Message
+	isCoseHashEnvelope bool // set to true if base represents a COSE hash envelope
```
shizhMSFT: Can we replace it with a method of `envelope`?

signature/types.go
ghost: Updated.

signature/cose/envelope.go
ghost: Updated.

signature/cose/envelope.go
```diff
@@ -297,27 +311,71 @@ func (e *envelope) Verify() (*signature.EnvelopeContent, error) {
...
+			}
+			coseHashEnvelopPayload.Location = payloadLocation.(string)
+		}
+		return &signature.EnvelopeContent{
+			SignerInfo: *signerInfo,
+			Payload: signature.Payload{
+				CoseHashEnvelopePayload: &coseHashEnvelopPayload,
+			},
+		}, nil
+	}
```
JeyJeyGao: Is it better to move this section into`e.payload()`?

ghost: Updated.

signature/types.go
```diff
@@ -182,15 +186,31 @@ type SignerInfo struct {
...
-// Payload represents payload in bytes and its content type.
+// Payload is the payload to be signed.
+//
+// When CoseHashEnvelopePayload is provided, ContentType and Content are ignored.
+// In this case, the payload represents a COSE Hash Envelope payload.
+// Reference: https://www.ietf.org/archive/id/draft-ietf-cose-hash-envelope-05.html
 type Payload struct {
 	// ContentType specifies the content type of payload.
+	//
+	// It is ignored, when [Payload.CoseHashEnvelopePayload] is present.
```
JeyJeyGao: Why the ContentType was ignored when using CoseHashEnvelopePayload? Do we need to define a new payload type for cose hash envelope?

ghost: Please take a look at https://www.ietf.org/archive/id/draft-ietf-cose-hash-envelope-05.html for details.
In Cose Hash Envelope, ContentType is not used. `preimage-content-type` shall be used.

signature/internal/base/envelope.go
```diff
@@ -190,10 +175,9 @@ func validateSigningAndExpiryTime(signingTime, expireTime time.Time) error {
 
 // validatePayload performs validation of the payload.
 func validatePayload(payload *signature.Payload) error {
-	if len(payload.Content) == 0 {
-		return errors.New("content not present")
+	if payload.CoseHashEnvelopePayload == nil && len(payload.Content) == 0 {
+		return errors.New("at least one of payload.Content and payload.CoseHashEnvelopePayload is required")
```
JeyJeyGao:
```suggestion
		return errors.New("one of payload.Content and payload.CoseHashEnvelopePayload is required")
```

JeyJeyGao (dismissed):
LGTM with a suggestion

signature/types.go
```diff
@@ -182,15 +186,31 @@ type SignerInfo struct {
...
 	Content []byte
+
+	// CoseHashEnvelopePayload is the payload to be signed when signing
+	// under the COSE format and COSE Hash Envelope is required as result.
+	//
+	// When present, [Payload.ContentType] and [Payload.Content] are ignored.
+	//
+	// Reference: https://www.ietf.org/archive/id/draft-ietf-cose-hash-envelope-05.html
+	CoseHashEnvelopePayload *cose.HashEnvelopePayload
```
shizhMSFT: nit: code style: naming convention
```suggestion
	COSEHashEnvelopePayload *cose.HashEnvelopePayload
```

signature/cose/envelope.go
```diff
@@ -390,10 +440,18 @@ func (e *envelope) signerInfo() (*signature.SignerInfo, error) {
...
 	return &signerInfo, nil
 }
 
+// isCoseHashEnvelope returns true if e.base is a COSE hash envelope.
+//
+// Reference: https://www.ietf.org/archive/id/draft-ietf-cose-hash-envelope-05.html
+func (e *envelope) isCoseHashEnvelope() bool {
+	_, ok1 := e.base.Headers.Protected[cose.HeaderLabelPayloadHashAlgorithm]
+	_, ok2 := e.base.Headers.Protected[cose.HeaderLabelContentType]
+	return ok1 && !ok2
```
shizhMSFT: code style: use meaningful variable names

signature/cose/envelope.go
```diff
@@ -227,13 +227,25 @@ func (e *envelope) Sign(req *signature.SignRequest) ([]byte, error) {
...
+		hashEnv, err := cose.SignHashEnvelope(rand.Reader, signer, msg.Headers, *req.Payload.CoseHashEnvelopePayload)
+		if err != nil {
+			return nil, &signature.InvalidSignatureError{Msg: err.Error()}
+		}
+		if err := hashEnvMsg.UnmarshalCBOR(hashEnv); err != nil {
+			return nil, &signature.InvalidSignatureError{Msg: err.Error()}
+		}
+		msg.Payload = hashEnvMsg.Payload
+		msg.Signature = hashEnvMsg.Signature
+		mergeCoseHashEnvelopeProtectedHeader(msg.Headers.Protected, hashEnvMsg.Headers.Protected)
```
shizhMSFT, line 247: Just curious why we can't just do `msg = hashEnvMsg` and then set `msg.Headers.Raw*` to `nil`?

signature/cose/envelope.go
```diff
@@ -211,6 +209,8 @@ func ParseEnvelope(envelopeBytes []byte) (signature.Envelope, error) {
 
 // Sign implements signature.Envelope interface.
 // On success, this function returns the COSE signature envelope byte slice.
+// When req.Payload.CoseHashEnvelope is present, it returns a COSE hash envelope
+// byte slice.
 func (e *envelope) Sign(req *signature.SignRequest) ([]byte, error) {
```
shizhMSFT, line 214: code style: it seems we can refactor line 222 - 237 out to have something like
```go
func signCOSESign1(s signer, req *signature.SignRequest) (*cose.Sign1Message, error) {
```
then we can have something like
```go
func signCOSEHashEnvelope(s signer, req *signature.SignRequest) (*cose.Sign1Message, error) {
```

and do something like
```go
var msg *cose.Sign1Message
if req.Payload.CoseHashEnvelopePayload != nil { // COSE hash envelope
	msg, err = signCOSEHashEnvelope(signer, req)
} else {
	msg, err = signCOSESign1(signer, req)
}
if err != nil {
	return nil, err
}
```

signature/cose/envelope.go
```diff
@@ -329,6 +353,32 @@ func (e *envelope) Content() (*signature.EnvelopeContent, error) {
 
 // Given a COSE envelope, extracts its signature.Payload.
 func (e *envelope) payload() (*signature.Payload, error) {
```
shizhMSFT, line 360: This function is too large. Can we refactor it to smaller functions?

ghost: This was a suggestion from https://github.com/notaryproject/notation-core-go/pull/268#discussion_r2055184535. 
Let me improve it further.

signature/cose/envelope.go
ghost: I think the previous style is good enough though. Because to sign, we need to generate the protected header first. That means, in the above suggestion, we will need to add the `generate protected header` code in both `signCOSEHashEnvelope` and `signCOSESign1`, which is not as straightforward as the current code flow.

signature/cose/envelope.go
ghost: This is indeed very subtle. After `UnmarshalCBOR`, the time in hashEnvMsg.Headers.Protected are of type `time.Time`. However, for signing, we require the time type to be `cbor.RawMessage` representing a Tag1 Datetime CBOR object. It is checked after signing, therefore, directly using the unmarshalled `Sign1Message` would fail. 
Reference: 
https://github.com/notaryproject/notation-core-go/blob/ef8789627fc3c764e1563a2999b33381e7992963/signature/cose/envelope.go#L676
https://github.com/notaryproject/specifications/pull/199#discussion_r1013748433

signature/cose/envelope.go
ghost: Updated.

signature/cose/envelope.go
ghost: Updated.

signature/types.go
ghost: Updated.

signature/internal/base/envelope.go
ghost: Updated.

signature/cose/envelope.go
shizhMSFT: Can you add a comment block for this?

signature/cose/envelope.go
shizhMSFT: It's on your call.

JeyJeyGao (dismissed):
LGTM

signature/cose/envelope.go
ghost: Added.

shizhMSFT (approved):
LGTM

JeyJeyGao (approved):
LGTM

PR #269, opened 2025-04-15
fix: error message of `SignatureAuthenticityError`

### LINKED ISSUE #266: Return more details in `SignatureAuthenticityError` from `VerifyAuthenticity`
### Is your feature request related to a problem?

https://github.com/notaryproject/notation-core-go/blob/4d7353253409c34db464cf4ca2802d6f218b95fd/signature/signer.go#L146
The `VerifyAuthenticity` method will return an empty `SignatureAuthenticityError` which will directly returned by `notation-go` in https://github.com/notaryproject/notation-go/blob/3bd0ac92b2bad47e477723bdb16a968a089738ae/verifier/verifier.go#L740.
The library caller like ratify will only get an empty error without any details, which makes it difficult to render the error to users.

### What solution do you propose?

Can we add some details to this error which can be propagated to downstreams.

### What alternatives have you considered?

n/a

### Any additional context?

_No response_

ghost:
Resolves #266

### DIFF
--- a/signature/errors.go
+++ b/signature/errors.go
@@ -67,7 +67,7 @@ type SignatureAuthenticityError struct{}
 
 // Error returns the default error message.
 func (e *SignatureAuthenticityError) Error() string {
-	return "signature is not produced by a trusted signer"
+	return "the signature's certificate chain does not cotain any trusted certificate"
 }
 
 // UnsupportedSigningKeyError is used when a signing key is not supported.
--- a/signature/errors.go
+++ b/signature/errors.go
@@ -67,7 +67,7 @@ type SignatureAuthenticityError struct{}
 
 // Error returns the default error message.
 func (e *SignatureAuthenticityError) Error() string {
-	return "the signature's certificate chain does not cotain any trusted certificate"
+	return "the signature's certificate chain does not contain any trusted certificate"
 }
 
 // UnsupportedSigningKeyError is used when a signing key is not supported.
--- a/signature/errors_test.go
+++ b/signature/errors_test.go
@@ -162,7 +162,7 @@ func TestSignatureEnvelopeNotFoundError(t *testing.T) {
 
 func TestSignatureAuthenticityError(t *testing.T) {
 	err := &SignatureAuthenticityError{}
-	expectMsg := "signature is not produced by a trusted signer"
+	expectMsg := "the signature's certificate chain does not cotain any trusted certificate"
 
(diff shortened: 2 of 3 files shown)


signature/signer.go
```diff
@@ -122,9 +122,10 @@ func (s *localSigner) PrivateKey() crypto.PrivateKey {
 	return s.key
 }
 
-// VerifyAuthenticity verifies the certificate chain in the given SignerInfo
-// with one of the trusted certificates and returns a certificate that matches
-// with one of the certificates in the SignerInfo.
+// VerifyAuthenticity iterates the certificate chain in signerInfo, for each
```
JeyJeyGao, line 125: nit: to link the type
```suggestion
// VerifyAuthenticity iterates the certificate chain in [signerInfo], for each
```

shizhMSFT (approved):
LGTM

signature/errors.go
```diff
@@ -67,7 +67,7 @@ type SignatureAuthenticityError struct{}
 
 // Error returns the default error message.
 func (e *SignatureAuthenticityError) Error() string {
-	return "signature is not produced by a trusted signer"
+	return "the signature's certificate chain does not contain any trusted certificate"
```
yizha1, line 70:
```suggestion
	return "the certificate chain for the signature does not contain any trusted identities as per the trust policy configuration"
```

signature/errors_test.go
```diff
@@ -162,7 +162,7 @@ func TestSignatureEnvelopeNotFoundError(t *testing.T) {
 
 func TestSignatureAuthenticityError(t *testing.T) {
 	err := &SignatureAuthenticityError{}
-	expectMsg := "signature is not produced by a trusted signer"
+	expectMsg := "the signature's certificate chain does not contain any trusted certificate"
```
yizha1, line 165:
```suggestion
	expectMsg := "the certificate chain for the signature does not contain any trusted identities as per the trust policy configuration"
```

signature/errors.go
Two-Hearts: @yizha1 This is notation-core-go, the error message is not related to trust policy or even trust store. The function that returns this error just check the signature's cert chain against a list of trusted certificates.

JeyJeyGao (approved):
LGTM

yizha1 (approved):
LGTM, but need a PR in notation-go to return more actionable message to the user/caller

PR #272, opened 2025-04-21
chore: update workflow for feature/cose-hash-envelope branch

ghost:
This PR targets on `feature/cose-hash-envelope` branch.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,9 +15,13 @@ name: Build
 
 on:
   push:
-    branches: main
+    branches: 
+      - main
+      - cose*
   pull_request:
-    branches: main
+    branches: 
+      - main
+      - cose*
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -15,9 +15,13 @@ name: Build
 
 on:
   push:
-    branches: main
+    branches: 
+      - main
+      - cose*
```
shizhMSFT: The branch name is not correct.

shizhMSFT (changes requested):

.github/workflows/build.yml
ghost: Updated.

.github/workflows/build.yml
```diff
@@ -15,9 +15,13 @@ name: Build
 
 on:
   push:
-    branches: main
+    branches: 
+      - main
+      - feat*
```
shizhMSFT: Shouldn't it be something like `feature/*`?

shizhMSFT (changes requested):

shizhMSFT (approved):
LGTM

JeyJeyGao (approved):
LGTM

PR #276, opened 2025-05-07
add govulncheck workflows for scanning CVEs

qba73:
This PR adds [govulncheck](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck) security scanning workflows


qba73:
This PR covers: https://github.com/notaryproject/notation-go/issues/534

.github/workflows/reusable-security-audit.yml
JeyJeyGao, line 1: Do we really need to split them to 2 files?

JeyJeyGao, line 1: This workflow is for vulnerability scanning, which is different from a security audit. For example, here is our actual security audit: https://github.com/notaryproject/specifications/tree/main/security/reports/audit. So we may need a better name for the file.

qba73: updated

qba73: applied

JeyJeyGao (dismissed):
LGTM

